PR #1, opened 2020-07-30
Add stricter type checking to the library

mariusandra:
Previously the call to `posthog.identify(distinctId, properties)` required a string `distinctId` argument, even though `null` is perfectly acceptable if you just want to set properties.

Typescript was not in `strict` mode, thus the `| null` part in `string | null` was omitted from the generated definitions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,12 +4,11 @@
     "description": "Reimplementation of posthog-js to be as light and modular as possible.",
     "main": "dist/src/index.js",
     "scripts": {
-        "start": "npm run build",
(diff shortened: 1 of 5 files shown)


Twixes:
Looks good. I just replaced npm with yarn in scripts (since yarn.lock is in use) and made "API" uppercase in type name (I think it looks better, but you can revert if you're committed to "Api").

mariusandra:
Thanks! Looks good to me!
I'll merge and release a new version.

PR #4, opened 2022-07-12
React Native "native" SDK 

benjackwhite:
> Note: Eventually this will be merged into `posthog-js` but for now it is here as in spirit it is trying to be a "lite" version of the RN library as well as a starting point for a unified JS core.

# What do I want to actually release?

* `/examples/example-expo` > An example Expo app showing how Posthog can be used without any native rebuilding
* `/posthog-core` > The "core" package containing all non-platform specific code (interface, queueing logic etc.)
* `/posthog-react-native` > The new RN SDK that uses the core as its base and adds new functionality as well as removing the dependency on native code (i.e. posthog-ios/android)
   * Features
   - [x] No native dependencies needed
   - [x] Autocapture - Touch events are tracked similar to the web with special React props detected `ph-label` as well as not-tracking with `ph-no-capture`
   - [x] `useLifecycleTracking` > React hook to take care of correctly tracking events related to opening, backgrounding etc.
   - [x] `useNavigationTracker` > Automatic Screen tracking of routes when using `react-navigation/native` with controls for customising the screen name or properties tracked
   - [x] Feature flags
   - [x] Groups
   - [x] Sessions

## What is more of an experiment
* `/posthog-web` > Actually moving the `posthog-js` to this unified core is of course a massive undertaking. This is more of an example to show how it could work and to make sure the design of the `core` is not limiting a browser platform
* `/posthog-node` > This is more realistic that we could migrate over to a shared core but is not fully functioning. Most crucial would be handling the fact that this SDK will be used for multiple "users" rather than one persisted user. Nonetheless it shows high-potential

## Outstanding TODOs

- ❌ Ideally, the new RN lib would work with the old one such that persisted data like `distinct_id` is re-used. I started on this for iOS but an Android workaround would be needed and this would need some way of testing it works
  - iOS should work but Android uses a completely non-expo supported way of storing data
- [x] Improve the build system - I'm using yarn workspaces (and yalc for the example app) but each package should be built with minified bundles, js, types etc. It's not perfect yet.
- [x] Get the package version from the package.json - couldn't get this to work without weird tsconfig issues
- [x] Ensure Opt out / in logic is working perhaps with persistence?
- [x] Enable retries
- ❌ Ensure backwards compatible with current React Native library
  - Too many changes here. We can at least keep the overall API the same, but instantiation is simply different now
- [x] Feature flags 
  - [x] polling 
  - [x] subscribing 
  - [x] overriding
  - [x] persisting to storage
  - [x] adding to tracking events
- [x] Decide on the validation code. It was mostly copied from `posthog-node` and its questionable how useful it is.
- [x] Save the groups information properly to be used for feature flags
- [x] Persist the queued events to some sort of storage. Especially for mobile this is important as apps can be killed randomly.
- [x] Reset
- [x] Listener for events (feature flags, flush, queue)
- [x] Tests tests tests tests...
- [x] Unref timers for Nodejs?
- [x] React Native E2E tests (Detox could be an option)
- [x] CI / CD flows
- [x] This: https://github.com/PostHog/posthog-js/pull/404/files

Follow up

- [ ] Detect network errors and either not flush or add back to queue
- [x] SessionID tracking
- [x] Mask autocapture fields that could be sensitive - all props
- [x] Pull out autocapture into a separate PR


posthog-core/src/index.ts
```diff
@@ -0,0 +1,460 @@
...
+    })
+  }
+
+  flush(callback?: (err?: any, data?: any) => void) {
+    if (!this.enabled) {
+      return callback && this.setImmediate(callback)
+    }
+
+    if (this._flushTimer) {
+      clearTimeout(this._flushTimer)
```
weyert, line 384:
```
this._flushTimer.unref()
```

posthog-core/src/eventemitter.ts
```diff
@@ -0,0 +1,27 @@
+export class SimpleEventEmitter {
+  events: { [key: string]: ((e: any) => void)[] } = {}
+
+  constructor() {
```
weyert, line 4: Consider introducing/adding a `shutdown` or similar named function which can be used to clean up of things like:

- Call `flush` so things get flushed
- Stop any running timers, e.g. setInterval / setTimeout etc  
- Ensure `unref()` is called on timers, see also: https://httptoolkit.tech/blog/unblocking-node-with-unref/
- Abort  any on-going requests (if that makes sense?)

posthog-core/src/index.ts
```diff
@@ -0,0 +1,460 @@
...
+
+  // internal
+  private _events: SimpleEventEmitter
+  private _queue: PostHogQueueItem[]
+  private _flushTimer?: any
+
+  private _decideResponsePromise?: Promise<PostHogDecideResponse>
+  private _decideResponse?: PostHogDecideResponse
+  private _decideTimer?: any
+  private _decidePollInterval = 10000
```
weyert, line 34: Looks like we can't override this interval via `PosthogCoreOptions`?

posthog-core/src/index.ts
```diff
@@ -0,0 +1,460 @@
...
+        [key: string]: any
+      }
+    | undefined
+
+  enable() {
+    this.enabled = true
+  }
+
+  disable() {
+    this.enabled = false
```
weyert, line 99: Shouldn't this stop any timers? E.g. flush interval or decide interval?

weyert (commented):
Looking great! I have might some suggestions that I saw while having a quick look :)

posthog-core/src/eventemitter.ts
benjackwhite: Yep! Sounds good and what I'm hoping todo (I want separate tests for the nodejs side of things as things like unref are unique to that platform

posthog-core/src/index.ts
benjackwhite: This is only temporarily hardcoded, I'll be adding it to the options for sure 👍

posthog-core/src/index.ts
benjackwhite: Yeah, this is not finished yet. The web side of disabling / enabling is so much more complicated so I'm leaving this till the rest of the lib is stable to see how much crossover there is between the platform 👍

benjackwhite:
Reminder to post [here](https://github.com/PostHog/posthog/issues/1174) once we have this in a testable state

rollup.config.js
```diff
@@ -0,0 +1,70 @@
...
+import typescript from 'rollup-plugin-typescript2'
+import dts from 'rollup-plugin-dts'
+
+import pkg from './package.json'
+
+const extensions = ['.js', '.jsx', '.ts', '.tsx']
+
+let globalExternal = Object.keys(pkg.dependencies || {}).concat(Object.keys(pkg.peerDependencies || {}))
+
+const configs = ['posthog-react-native', 'posthog-node', 'posthog-web'].reduce((acc, x) => {
```
weyert, line 14: You might find tsup interesting too. Not sure how well it handles React Native

posthog-core/src/index.ts
```diff
@@ -0,0 +1,554 @@
...
+    const featureFlags = this.getFeatureFlags()
+
+    if (!featureFlags) {
+      // If we haven't loaded flags yet we respond undefined to indicate this
+      return undefined
+    }
+
+    if (this.sendFeatureFlagEvent && !this.flagCallReported[key]) {
+      this.flagCallReported[key] = true
+      this.capture('$feature_flag_called', {
```
liyiy, line 350: This is how it also works in posthog-js but @neilkakkar how come we're sending this capture call in the server libraries regardless of whether we're sending `sendFeatureFlag` but in the client side we only do this under the condition?

liyiy: And we don't have to send the feature flag properties in the `capture` call here

benjackwhite: > And we don't have to send the feature flag properties in the `capture` call here

I don't quite follow what you mean by this?

neilkakkar: >  @neilkakkar how come we're sending this capture call in the server libraries regardless of whether we're sending sendFeatureFlag but in the client side we only do this under the condition?

good catch, I wasn't aware that posthog-js doesn't send it with every call, but once per user per library instantiation.

neilkakkar: Since it doesn't affect experiments whether it's once or more than once per user, I guess makes sense for server-side to follow the same.

liyiy: @benjackwhite I think you're good here! Just some confusion I had with server vs client side implementation of the feature flag called 😅

posthog-core/src/index.ts
```diff
@@ -0,0 +1,556 @@
...
+      properties: {
+        ...this.props, // Persisted properties first
+        ...(payload.properties || {}), // Followed by user specified properties
+        ...this.getCommonEventProperties(), // Followed by common PH props
+        $session_id: this.getSessionId(),
+      },
+    }
+  }
+
+  getSessionId(): string | undefined {
```
liyiy, line 142: I think we're missing a `resetSessionId()` method based on https://github.com/PostHog/posthog/issues/10780

benjackwhite: Good catch, I'll add it

PR #6, opened 2022-07-28
Fixed anonymous id calls

benjackwhite:
@neilkakkar quick eyes on this would be great just to roughly see that it looks sensible

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -124,7 +124,6 @@ export abstract class PostHogCore {
     for (const key in PostHogPersistedProperty) {
       this.setPersistedProperty((PostHogPersistedProperty as any)[key], null)
     }
-    this.setPersistedProperty(PostHogPersistedProperty.DistinctId, generateUUID(globalThis))
   }
 
   debug(enabled: boolean = true): void {
@@ -164,13 +163,17 @@ export abstract class PostHogCore {
     this.setPersistedProperty(PostHogPersistedProperty.SessionId, null)
   }
 
-  getDistinctId(): string {
-    let distinctId = this.getPersistedProperty<string>(PostHogPersistedProperty.DistinctId)
-    if (!distinctId) {
-      distinctId = generateUUID(globalThis)
-      this.setPersistedProperty(PostHogPersistedProperty.DistinctId, distinctId)
+  getAnonymousId(): string {
+    let anonId = this.getPersistedProperty<string>(PostHogPersistedProperty.AnonymousId)
+    if (!anonId) {
+      anonId = generateUUID(globalThis)
+      this.setPersistedProperty(PostHogPersistedProperty.AnonymousId, anonId)
     }
-    return distinctId
+    return anonId
+  }
+
+  getDistinctId(): string {
+    return this.getPersistedProperty<string>(PostHogPersistedProperty.DistinctId) || this.getAnonymousId()
   }
 
   register(properties: { [key: string]: any }): void {
@@ -190,7 +193,8 @@ export abstract class PostHogCore {
    *** TRACKING
    ***/
   identify(distinctId?: string, properties?: PostHogEventProperties): this {
-    distinctId = distinctId || this.getDistinctId()
+    const previousDistinctId = this.getDistinctId()
+    distinctId = distinctId || previousDistinctId
 
     if (properties?.$groups) {
       this.groups(properties.$groups)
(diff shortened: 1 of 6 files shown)


posthog-core/src/index.ts
```diff
@@ -202,27 +206,27 @@ export abstract class PostHogCore {
...
     return this
   }
 
   capture(event: string, properties?: { [key: string]: any }): this {
-    // NOTE: Legacy nodejs implementation uses groups
-    if (properties && properties['groups']) {
-      properties.$groups = properties.groups
-      delete properties.groups
-    }
-
```
benjackwhite, line 225: This was meant to be removed

posthog-core/src/index.ts
```diff
@@ -202,27 +206,27 @@ export abstract class PostHogCore {
         event: '$identify',
         properties: {
           ...(properties || {}),
-          $anon_distinct_id: this.getDistinctId(), // TODO: Should this be the previous distinct id or the original anon id??
+          $anon_distinct_id: this.getAnonymousId(),
```
neilkakkar, line 209: This should always be the previous distinctID?

As we talked earlier, if people forget to set `reset()` and we call `identify()` with the anon_distinct_id set to an anon ID, then both the distinctIDs supplied by the user will be merged, which isn't something we want.

If anon_distinct_id is an existing identified id though, they're not merged together (you need an `alias` call to do this explicitly).

posthog-core/src/index.ts
```diff
@@ -202,27 +206,27 @@ export abstract class PostHogCore {
...
         },
       }),
       $set: properties,
     }
 
-    if (distinctId !== this.getDistinctId()) {
+    if (distinctId !== previousDistinctId) {
+      // We keep the AnonymousId to be used by decide calls and identify to link the previousId
+      this.setPersistedProperty(PostHogPersistedProperty.AnonymousId, previousDistinctId)
```
neilkakkar, line 217: .. oh, this is sortof confusing. This makes the anonymousId be a non-anonymous ID ? (and makes my above comment moot)

benjackwhite: Yeah its how the posthog-js seems to do it. Just overwrites the anonymousId. That way I guess there is _never_ the chance that it accidentally gets used

posthog-core/src/index.ts
benjackwhite: As you saw in the other comment, this is exactly what essentially happens as the anonId is overwritten with the old distinctId.

posthog-core/src/index.ts
```diff
@@ -331,7 +335,12 @@ export abstract class PostHogCore {
     const fetchOptions: PostHogFetchOptions = {
       method: 'POST',
       headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify({ groups, distinct_id: distinctId, token: this.apiKey }),
+      body: JSON.stringify({
+        token: this.apiKey,
+        distinct_id: distinctId,
+        $anon_distinct_id: this.getAnonymousId(),
```
neilkakkar, line 341: this works, although note that we don't need to send this everytime someone reloads `/decide`. Atleast once after an `identify` event is sufficient.

I follow this in posthog-js (and set the anon ID to be sent to `undefined` afterwards).

The reason being that sending anon_id when it doesn't need to be can have a perf impact.

.. .Thinking about this a bit further, we're fine to not send these whenever `distinctId === this.getAnonymousId()`, and send them whenever they are different, since that implies an identify call happened right before.

Noice, I like this simplification.

neilkakkar (commented):
Looking mostly good!

Any recommendations for things I should change in the docs?

posthog-core/src/index.ts
benjackwhite: I think the "only need to send this once" is the tricky / confusing part. It's much harder to make it only get sent once as that requires adding extra code to clear it _after_ the decide call. If it is genuinely required I think the docs could do with saying exactly that :D 

If this is the _only_ use case for keeping the previous distinct ID around then I guess we are safe to just call `setAnonId(null)` after the decide response comes back. But then on reload the /decide call wouldn't get the anon_id which means the flag would potentially come wrong, right??

neilkakkar: > But then on reload the /decide call wouldn't get the anon_id which means the flag would potentially come wrong, right??

Nope, so basically whenever anon_id is sent alongside, it writes to a table to tell us to always associate this user with this given id. Any subsequent calls, with or without the anonID, give the 'right' flags.

---

Agree about it being the tricky part. Hmm.. Actually, yeah, let's keep clients simple. I can tweak the server side to remain performant by following the heuristic I mentioned above.

benjackwhite: Awesome then that makes total sense.

PR #14, opened 2022-08-09
feat(feature-flags): Enable local evaluation

### LINKED ISSUE #13: Feature flag schema issues
The new version seems to send a new `$enabled_feature_flags` property with events, which isn't used elsewhere.

Instead, what we'd expect to see are `$feature/<feature-key>` properties where the value is the value of the flag.

neilkakkar:
Following spec from https://github.com/PostHog/posthog/issues/10459

Also fixes #13 

And fixes some more tricky bugs I encountered:

1. Node library was dropping most events, as the `Queue` is cleared on every new call to the library. Not greaat 😬 .
2. lazy loading props wasn't working properly, which meant groups belonging to one event would get added to another event. 😬 
3. Still open: Figuring out what timer/promise is causing an unclean shutdown when awaiting flush.

Other than this, feature flags are mostly sorted: things work as expected, tested locally 🙌


posthog-core/src/index.ts
```diff
@@ -491,7 +562,10 @@ export abstract class PostHogCore {
     const queue = this.getPersistedProperty<PostHogQueueItem[]>(PostHogPersistedProperty.Queue) || []
 
     if (!queue.length) {
-      return callback && safeSetTimeout(callback, 0)
+      callback?.()
+      // TODO: Ask Ben what timeout on the callback really does. I don't understand
```
neilkakkar, line 566: Hey @benjackwhite I'm a bit mystified, not sure what's going on here:

1. Things are fine when I just run `example.ts`
2. When running jest tests though, if I wait for `safeSetTimeout`, some tests hang! (You can test this by commenting out the `callback?.()` call.

I'm not quite sure why this is happening. Do you have any ideas? I tried process dumps to figure out any unknown timers / promises that might be stopping things, but didn't manage to figure this out 😅 .

posthog-core/src/index.ts
```diff
@@ -268,9 +280,9 @@ export abstract class PostHogCore {
 
   groups(groups: { [type: string]: string | number }): this {
     // Get persisted groups
-    const existingGroups = this.props.$groups || {}
+    const existingGroups =
```
neilkakkar, line 283: also cc: @benjackwhite , the `reInit` was clearing persisted props, but this was sneaky, just using `this.props` directly.

Have replaced it in 3 places. Don't think it should affect anything else, and tests all pass, but have a look if you think it will?

neilkakkar: hmm, it's a bit confusing, I'm not quite sure why `this.props` exists. Can do away with it..

neilkakkar: Ah, I see, its the lazy loading getter which creates problems, as we can't tell when its been invalidated behind us.

posthog-core/src/index.ts
```diff
@@ -313,30 +332,77 @@ export abstract class PostHogCore {
...
+
+    this.setPersistedProperty<PostHogEventProperties>(PostHogPersistedProperty.PersonProperties, {
+      ...existingProperties,
+      ...properties,
+    })
+
+    return this
+  }
+
+  // Test heavilyy!
```
neilkakkar, line 353: I should test this separately too!

posthog-core/src/index.ts
```diff
@@ -436,6 +502,15 @@ export abstract class PostHogCore {
     return this.setPersistedProperty(PostHogPersistedProperty.OverrideFeatureFlags, flags)
   }
 
+  _sendFeatureFlags(event: string, properties?: { [key: string]: any }): void {
+    // Used for posthog-node only
```
benjackwhite, line 506: If this is node only, why do we want it in the core? Why not have this called by the already separate capture call?

posthog-core/src/index.ts
```diff
@@ -363,8 +429,8 @@ export abstract class PostHogCore {
     const featureFlags = this.getFeatureFlags()
 
     if (!featureFlags) {
-      // If we haven't loaded flags yet we respond undefined to indicate this
-      return undefined
+      // If we haven't loaded flags yet, or errored out, we respond with the default result
+      return defaultResult
```
benjackwhite, line 433: This was intentionally left as undefined. Otherwise there is no way for the client to know whether they are getting the actual loaded flag or just the default.

Not sure what the benefit of returning the defaultResult is given that the client could just do `posthog.getFeatureFlag(foo) ?? myDefaultResult`.

posthog-core/src/index.ts
```diff
@@ -375,8 +441,8 @@ export abstract class PostHogCore {
       })
     }
 
-    // If we have flags we either return the value (true or string) or the defaultResult
-    return featureFlags[key] ?? defaultResult
+    // If we have flags we either return the value (true or string) or undefined
```
benjackwhite, line 444: This really confuses me 😅 

So if I call `posthog.getFeatureFlag("foo", false)` I should expect the response to be:
* `false` if the flags aren't loaded yet
* `true` if the flags are loaded and the value is true
* `undefined` if the flags are loaded but there is no flag value

This seems the opposite of what I would expect using this method.

posthog-core/src/index.ts
benjackwhite: tbh I think we can remove the safeSetTimeout here as we immediately return anyways. It was only necessary for the old API where we had lots of callbacks in a loop.

Not sure why the tests have an issue though. I guess becuase there is a hanging timeout ref if you haven't run `runAllPendingTimers` but that should be covered by the unref part of the `safeSetTimeout` function...

posthog-node/src/posthog-node.ts
```diff
@@ -80,12 +113,12 @@ export class PostHogGlobal implements PostHogNodeV1 {
...
   }
 
-  capture({ distinctId, event, properties, groups }: EventMessageV1): void {
+  capture({ distinctId, event, properties, groups, sendFeatureFlags }: EventMessageV1): void {
     this.reInit(distinctId)
     if (groups) {
       this._sharedClient.groups(groups)
     }
-    this._sharedClient.capture(event, properties)
+    this._sharedClient.capture(event, properties, sendFeatureFlags || false)
```
benjackwhite, line 121: This functionality I think should exist only here if it is truly node specific.

posthog-node/src/posthog-node.ts
```diff
@@ -101,39 +134,153 @@ export class PostHogGlobal implements PostHogNodeV1 {
...
     key: string,
     distinctId: string,
-    groups?: Record<string, string> | undefined
+    defaultResult?: boolean | string,
+    groups?: Record<string, string>,
+    personProperties?: Record<string, string>,
+    groupProperties?: Record<string, Record<string, string>>,
+    onlyEvaluateLocally: boolean = false,
+    sendFeatureFlagEvents: boolean = true
   ): Promise<string | boolean | undefined> {
```
benjackwhite, line 143: Not sure about this API tbh. I wonder if it isn't worth providing an object interface, that way it is much easier to add / rework properties in the future. 8 arguments for a function is definitely too many...

e.g. 😅 
`posthog.getFeatureFlag("flag", "id, false, null, { foo: "bar}, null, false, false)`

versus
```
posthog.getFeatureFlag("flag", "id", false, {
  personProperties: { foo: "bar" },
  sendFeatureFlags: false
}
```

posthog-node/src/posthog-node.ts
```diff
@@ -101,39 +134,153 @@ export class PostHogGlobal implements PostHogNodeV1 {
...
     key: string,
     distinctId: string,
-    defaultResult?: boolean | undefined,
-    groups?: Record<string, string> | undefined
+    defaultResult?: boolean,
+    groups?: Record<string, string>,
+    personProperties?: Record<string, string>,
+    groupProperties?: Record<string, Record<string, string>>,
+    onlyEvaluateLocally: boolean = false,
+    sendFeatureFlagEvents: boolean = true
```
benjackwhite, line 208: Same comment here as above

benjackwhite (changes requested):
Some initial feedback on the API / implementation. Didn't look to closely at the test code etc as there was a lot of it 😅 will have another pass once you've responded to my comments 👍

posthog-core/src/index.ts
weyert: I already coded against this change 🙈

posthog-core/src/index.ts
neilkakkar: because it meant exposing the `enqueue` internals of core, which felt a bit more messy to me.

It's not that other libraries can't make use of this, it's just that _today_ it's used for node only. Hmm, actually, comment is confusing, will remove it.

posthog-core/src/index.ts
neilkakkar: Hmm, was thinking about this, and wonder if its best to remove the `defaultResult` from the API interface.

Exactly because of what you say -> the defaultResult in the API isn't doing much at all.

For where I'm coming from, consider the case where `defaultResult` is `true`, this messes up any place we had `|| defaultResult`, as it turns a false valid reply into true.

---

@weyert as in the existing change?

posthog-core/src/index.ts
neilkakkar: The way the `decide` endpoint works is that no value means the flag is set to `false`. 

This is exactly the opposite of what we want if we replace it by defaultResult, and `defaultResult` is true / some other variant string.

posthog-core/src/index.ts
neilkakkar: that unblocks, cool

posthog-node/src/posthog-node.ts
neilkakkar: Thanks, was thinking the same. Tests became a bit too illegible with the current thing. I so miss keyword args :'(

The reason I stuck around was that when you want to set `onlyEvaluateLocally` to true, it's a good idea to have atleast some groups/person properties / etc. set, and the current API structure forces you to think about those params.

Will fix.

posthog-core/src/index.ts
neilkakkar: ... Will change! Will remove defaultResult.. in all libraries actually, and let `undefined` be the case when we fail, and let the client figure out defaults on their own.

neilkakkar: Actually, since all libs are complete, will just do this:

Change the 3rd bullet to
> `false` if the flags are loaded but there is no flag value

and `posthog.getFeatureFlag("foo", 'xyz')` returns `false` if the flags are loaded but there is no flag value as well.

posthog-core/src/index.ts
benjackwhite: Weyert is trialing this lib using the alpha version hence his comment.

I think removing it makes total sense as we're basically replicating native functionality in our API for no good reason

posthog-node/src/posthog-node.ts
```diff
@@ -101,39 +134,160 @@ export class PostHogGlobal implements PostHogNodeV1 {
...
     key: string,
     distinctId: string,
-    groups?: Record<string, string> | undefined
+    options?: {
+      groups?: Record<string, string>
+      personProperties?: Record<string, string>
+      groupProperties?: Record<string, Record<string, string>>
+      onlyEvaluateLocally?: boolean
+      sendFeatureFlagEvents?: boolean
+    }
```
benjackwhite, line 143: ❤️

benjackwhite (approved):
Looks much better 👍 I more looked through the local evaluation tests than the implementation. Trusting they are comprehensive enough 😅

Nice work!

neilkakkar:
Yep indeed, the tests are replicated across all libraries :)

This library was trickier than the rest because of the interaction with `posthog-core` and sharing most things, and the base functionality having bugs. Basically wanted your eyes on the integration between those two, and the bugfixes I'm doing for node (see description). 

Thanks for taking the time to review!

posthog-core/src/index.ts
```diff
@@ -313,30 +331,77 @@ export abstract class PostHogCore {
     return this
   }
 
+  /***
+   * PROPERTIES
+   ***/
+
+  // TODO: Figure out resetting
```
neilkakkar, line 338: todo sorted, remove

posthog-core/src/index.ts
```diff
@@ -402,13 +473,16 @@ export abstract class PostHogCore {
...
 
-  isFeatureEnabled(key: string, defaultResult: boolean = false): boolean {
-    const flag = this.getFeatureFlag(key, defaultResult) ?? defaultResult
-    return !!flag
+  isFeatureEnabled(key: string): boolean | undefined {
+    const response = this.getFeatureFlag(key)
+    if (response === undefined) {
+      return undefined
+    }
+    return !!this.getFeatureFlag(key)
```
neilkakkar, line 481:
```suggestion
    return !!response
```

neilkakkar:
TODO: Configure logging so DEBUG outputs only when DEBUG is set.

posthog-core/src/index.ts
weyert: Yes, what @benjackwhite. I am trying out the alpha of the Node package and updated my OpenFeature provider to correctly indicate missing feature flag when `undefined` is returned.

neilkakkar: I see, sounds good, the behaviour is consistent then in this version.

neilkakkar:
Will setup logging in a follow up PR. Merging this one.

PR #16, opened 2022-08-23
Rename export for node to PostHog

mariusandra:
No more `PostHogGlobal`

### DIFF
--- a/examples/example-node/example.ts
+++ b/examples/example-node/example.ts
@@ -1,10 +1,10 @@
-import PostHog from 'posthog-node'
+import { PostHog } from 'posthog-node'
 // @ts-ignore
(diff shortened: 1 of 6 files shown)


benjackwhite (approved):
lgtm but didnt check locally

neilkakkar:
Tested locally: Works for `cjs`.

When a module though, we need to `import {PostHog} from 'posthog-node'`. I'm guessing something to do with default imports not existing for modules?

mariusandra:
What if we just kill the default export? It's bad practice anyway, and people are willing to change code when they upgrade to a new major version like 2.0.

neilkakkar:
Sounds good to me

PR #20, opened 2022-09-02
feat: React Native: Optional disable lifecycle methods

benjackwhite:
In relation to a Squeak question - adds the possibility to disable elements of autocapture

https://squeak.cloud/question/175


posthog-react-native/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "posthog-react-native",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
pauldambra, line 3: 2.1.0 new feature rather than a fix?

pauldambra (approved):
LGTM

posthog-react-native/package.json
benjackwhite: Fair point

Issue #21, opened 2022-09-06
Revert back to using node-fetch


yakkomajuri:
We're now using `undici` for our HTTP client, but unfortunately `undici.fetch` only works on Node >= 16.8.

This is mentioned in the docs, and you can see it in their code:

```js
if (nodeMajor > 16 || (nodeMajor === 16 && nodeMinor >= 8)) {
  let fetchImpl = null
  module.exports.fetch = async function fetch (resource) {
    if (!fetchImpl) {
      fetchImpl = require('./lib/fetch')
    }
    const dispatcher = (arguments[1] && arguments[1].dispatcher) || getGlobalDispatcher()
    return fetchImpl.apply(dispatcher, arguments)
  }
  // ...
}
```

I'd love if we could revert back to node-fetch (or something else) to unblock what I believe is the vast majority of our users, since I doubt many people are running Node versions above that.

More context: https://posthogusers.slack.com/archives/C01GLBKHKQT/p1662474954526869

Written at @benjackwhite's request - I'm happy to take this though if you're massively busy

weyert:
Is the problem you want to support Node v14 too? Maybe it’s a good idea to allow the consumer to decide which fetch package to use? 16.8 is already more than a year old and had bunch of (net and http related) security vulnerability fixes

yakkomajuri:
@weyert expecting the majority of our users to be not only on Node 16 but Node 16.8 is quite unreasonable in my opinion. 

We need to provide analytics infrastructure that is immediately useful to most. Notice how posthog-js targets very old browsers for example.

Allowing customers to specify their own `fetch` is a possibility but honestly I'd rather not do this at this stage. If something ever goes wrong with the library it becomes infinitely harder to debug and support.

PR #22, opened 2022-09-08
feat: Swap to axios as default fetch provider

benjackwhite:
Closes https://github.com/PostHog/posthog-js-lite/issues/21

### DIFF
--- a/posthog-node/package.json
+++ b/posthog-node/package.json
@@ -7,7 +7,7 @@
     "prepublish": "cd .. && yarn build"
   },
   "engines": {
-    "node": ">=10"
+    "node": ">=14.17.0"
   },
   "license": "MIT",
   "author": {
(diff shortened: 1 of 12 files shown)


posthog-node/package.json
```diff
@@ -7,7 +7,7 @@
     "prepublish": "cd .. && yarn build"
   },
   "engines": {
-    "node": ">=10"
+    "node": ">=14.17.0"
```
benjackwhite, line 10: This is the version that supports AbortController

neilkakkar (approved):
Looks good! I assume you tried running the node example and confirmed it worked?

(if not yet, I'll consider this a blocker to merging!)

benjackwhite:
> Looks good! I assume you tried running the node example and confirmed it worked?
> 
> (if not yet, I'll consider this a blocker to merging!)

Indeed I have!

PR #24, opened 2022-09-08
feat(feature-flags): Add bootstrap option for feature flags

neilkakkar:
Just like https://github.com/PostHog/posthog-js/pull/444 but for react-native.

@benjackwhite I could use some help figuring out this works for react-native.

Current issues I'm facing:

1. The CI doesn't seem to be testing `posthog-react-native` ?
2. How do I get the new test created specifically for react-native to run? Unfamiliar with expo, and it's complaining about those imports.
3. Can you help testing a 'live' version of this and confirming it works?

### DIFF
--- a/posthog-react-native/src/posthog-rn.ts
+++ b/posthog-react-native/src/posthog-rn.ts
@@ -39,7 +39,6 @@ export class PostHog extends PostHogCore {
 
     // It is possible that the old library was used so we try to get the legacy distinctID
     void preloadSemiAsyncStorage().then(() => {
-      // We run this extra here as the storage may have been reset...
       this.setupBootstrap(options)
 
       if (!SemiAsyncStorage.getItem(PostHogPersistedProperty.AnonymousId)) {
--- a/posthog-react-native/src/posthog-rn.ts
+++ b/posthog-react-native/src/posthog-rn.ts
@@ -61,7 +61,7 @@ export class PostHog extends PostHogCore {
   }
   setPersistedProperty<T>(key: PostHogPersistedProperty, value: T | null): void {
     if (this._persistence === 'memory') {
-      return this._memoryStorage.getProperty(key)
+      return this._memoryStorage.setProperty(key, value)
     }
     return value !== null ? SemiAsyncStorage.setItem(key, value) : SemiAsyncStorage.removeItem(key)
   }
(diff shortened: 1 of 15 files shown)


benjackwhite:
>     1. The CI doesn't seem to be testing `posthog-react-native` ?

It does - It's right after the standard tests https://github.com/PostHog/posthog-js-lite/blob/master/.github/workflows/ci.yml#L22

>     2. How do I get the new test created specifically for react-native to run? Unfamiliar with expo, and it's complaining about those imports.

Expo is a hard beast to test but for this we can stick to the unit tests. They have to be run separate to the other tests due to some specific tsconfig rules for React Native...

>     3. Can you help testing a 'live' version of this and confirming it works?

Sure - I'll check it out!

benjackwhite:
@neilkakkar trying to fix the tests but not sure what the logic should be.
Currently its doing this:

1. On load bootstrap the things
2. If feature flags are given they are set 👍 
3. Immediately after we reload the feature flags

My assumption is we _don't_ want to immediately reload the flags from the API if they are given. Is that correct?

neilkakkar:
> My assumption is we don't want to immediately reload the flags from the API if they are given. Is that correct?

No, this is okay!

The idea is that some flags are available on immediate load, while we wait for `/decide` or new properties to reflect in the feature flags. 
Some flags would be impossible to bootstrap, hence the immediate reload ensures we get those as soon as possible as well.

posthog-react-native/src/posthog-rn.ts
```diff
@@ -39,6 +39,9 @@ export class PostHog extends PostHogCore {
 
     // It is possible that the old library was used so we try to get the legacy distinctID
     void preloadSemiAsyncStorage().then(() => {
+      // We run this extra here as the storage may have been reset...
+      this.setupBootstrap(options)
```
neilkakkar, line 43: oh, interesting. I see, in react-native, storage is setup async, and we'll need this to be present before we can bootstrap FFs?

.. I guess except when using the `memory` option?

benjackwhite: Yeah pretty much. It's annoying but there isn't much way around it

benjackwhite:
@neilkakkar this looks good now and everything passes. Turns out memorystorage for RN was always broken anyway so this is a good catch...

Might be worth a quick glance for yourself over what I changed but if that's all good then I'd say we are good to go

neilkakkar:
Much appreciated, thanks! I just added a test for the async storage as well.

PR #25, opened 2022-09-22
Local eval date operators support

liyiy:
Add support for date operators in feature flags local evaluation, modeled after https://github.com/PostHog/posthog-python/pull/75

### DIFF
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -420,14 +420,14 @@ function isValidRegex(regex: string): boolean {
 function convertToDateTime(value: string | number | (string | number)[] | Date): Date {
   if (value instanceof Date) {
     return value
-  } else if (typeof value === 'string') {
-    const timestamp = Date.parse(value)
-    if (isNaN(timestamp) == false) {
-      return new Date(timestamp)
+  } else if (typeof value === 'string' || typeof value === 'number') {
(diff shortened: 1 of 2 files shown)


posthog-node/src/feature-flags.ts
```diff
@@ -409,4 +417,18 @@ function isValidRegex(regex: string): boolean {
   }
 }
 
+function convertToDateTime(value: string | number | (string | number)[] | Date): Date {
+  if (value instanceof Date) {
+    return value
+  } else if (typeof value === 'string') {
+    const timestamp = Date.parse(value)
+    if (isNaN(timestamp) == false) {
```
liyiy, line 425: `Invalid Date` does not throw as an error so we have to parse and check for its date string validity

posthog-node/src/feature-flags.ts
```diff
@@ -409,4 +417,18 @@ function isValidRegex(regex: string): boolean {
...
+      return new Date(timestamp)
+    }
+    throw new InconclusiveMatchError(`${value} is in an invalid date format`)
+  } else {
+    throw new InconclusiveMatchError(`The date provided ${value} must be a string or date object`)
```
EDsCODE, line 430: Number just throws an error always?

EDsCODE (approved):

benjackwhite (approved):

PR #31, opened 2022-11-08
feat: Make Expo optional

benjackwhite:


posthog-react-native/src/storage.ts
```diff
@@ -1,69 +1,69 @@
...
-  const uri = (FileSystem.documentDirectory || '') + POSTHOG_STORAGE_KEY
-  // If we change POSTHOG_STORAGE_VERSION then we should migrate the persisted data here
-  try {
-    const stringContent = await FileSystem.readAsStringAsync(uri)
-    return JSON.parse(stringContent).content
-  } catch (e) {
-    return {}
-  }
-}
+// NOTE: The core prefers a synchronous storage so we mimic this by pre-loading all keys
```
pauldambra, line 8: flyby nit: "a synchronous" or "asynchronous"
```suggestion
// NOTE: The core prefers asynchronous storage so we mimic this by pre-loading all keys
```

benjackwhite: haha I commited this and realised its incorrect. It's meant to be "a synchronous" (as in not-async).

.github/workflows/release.yml
```diff
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - master
-      - feat/multi-domain-support # TODO: Remove this once merged
```
mariusandra, line 8: This comment is a Schrödinger's comment within this PR.

mariusandra (approved):
Looks good when reading through the code. Didn't run it though.

The new packaging system makes sense as well. I thought for a while if we could expose a different and "just one thing" `posthog-expo` library to make the installation easier for users, but we don't want to be in the business of releasing an update every time there's a new `expo-file-system`, so this is likely the easiest way after all.

👍

posthog-react-native/src/storage.ts
pauldambra: Is it better then as "...the core prefers synchronous..." so it's more parseable

This is, of course, the most important thing to focus on in this PR :p

mariusandra (approved):

posthog-react-native/README.md
```diff
@@ -8,6 +8,140 @@ Specifically, the [React Native integration](https://www.posthog.com/docs/integr
 
 ### [Join our Slack community.](https://join.slack.com/t/posthogusers/shared_invite/enQtOTY0MzU5NjAwMDY3LTc2MWQ0OTZlNjhkODk3ZDI3NDVjMDE1YjgxY2I4ZjI4MzJhZmVmNjJkN2NmMGJmMzc2N2U3Yjc3ZjI5NGFlZDQ)
 
+## Installation
+
+### Expo Apps
+
+```sh
+expo install posthog-react-native expo-file-system expo-application expo-device expo-localization
+```
```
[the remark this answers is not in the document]
benjackwhite: With expo you always use `expo install`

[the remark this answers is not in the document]
mariusandra: In my defence, `expo` and `yarn` are both very similar 4 letter words

giphy-downsized-large

Issue #34, opened 2022-11-15
How to get feature flag variants in react-native?


MoSattler:
I would like to get feature-flag variants in react native. 

- `getFeatureFlags()` seems to only return an array of flag names (even though the typing indicates otherwise). 
- `getFeatureFlag()` does not seem to work at all and only returns `false`, no matter if a flag is set or not.

benjackwhite:
Feature flags are not necessarily available on initial load. We have a special hook for this which helps take care of reacting to when the feature flags are loaded via the api request to the decide endpoint. If you call these functions directly you probably want to ensure the flags have been succesfully loaded first.

Have you followed [the docs here](https://posthog.com/docs/integrate/client/react-native#feature-flags)?

neilkakkar:
Also, if you want flags to be instantly available, you can bootstrap them: https://posthog.com/docs/integrate/client/react-native#bootstrapping-flags

PR #38, opened 2022-11-21
fix(core): feature flag typing

MoSattler:
The results I got from `posthog.getFeatureFlags()` were not matching the typing. The typing indicated that the return type should be `{ [key: string]: string | boolean }`, but I only ever got `string[]`.

So I checked what the API is actually returning:
Screenshot 2022-11-21 at 12 55 33

And indeed the API was returning an array of strings. So I changed the type with this PR.

Also see [this](https://github.com/PostHog/posthog-js-lite/issues/34#issue-1449972311).

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -486,15 +486,9 @@ export abstract class PostHogCore {
       return flags
     }
 
-    flags = flags || {}
+    flags = flags || []
 
-    for (const key in overriddenFlags) {
-      if (!overriddenFlags[key]) {
-        delete flags[key]
-      } else {
-        flags[key] = overriddenFlags[key]
-      }
-    }
+    flags.filter(flag => overriddenFlags.includes(flag))
 
     return flags
   }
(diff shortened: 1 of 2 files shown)


neilkakkar:
Hey Mo, did you try creating multivariate feature flags?

neilkakkar:
Sorry, I mean the main PostHog app, not this library

neilkakkar:
For the request you're sending in the main screenshot, can I see the request body as well please?

(feel free to hide the api key)

---

(hard for me to reproduce for RN because I don't have expo/ios setup with me right now. _But_, we use the same codebase for posthog-node, which works fine)

neilkakkar:
Huh weird, I get the right stuff when I `curl` the above^.

Question: Are you using your **project API key** or a personal API key for the token?

neilkakkar:
I see the problem - it has to do with your proxy which doesn't pass the `v` parameter on to posthog.

For you to test & confirm:

1. If you put `app.posthog.com` in place of `REPLACE_ME`, it should work.
2. If you put `your domain proxy` I think it doesn't work

neilkakkar:
I see. Then indeed the `v=2` parameter isn't being forwarded.

PR #39, opened 2022-11-24
feat: Add timeout option for all calls

timgl:
This will auto timeout any calls after 10 seconds (which may be too aggressive?). People were running into issues when we had slow API calls as it'd wait forever.

### DIFF
--- a/posthog-node/package.json
+++ b/posthog-node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "posthog-node",
-  "version": "2.2.0",
+  "version": "2.2.1",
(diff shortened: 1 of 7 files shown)


posthog-node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "posthog-node",
-  "version": "2.2.0",
+  "version": "2.2.1",
```
neilkakkar, line 3: if you're bumping version, add to changelog as well please!

neilkakkar (approved):

PR #43, opened 2022-12-07
feat: Allow custom timestamps

benjackwhite:
Adds a 3rd param to capture calls "options" which is an object of additional, more custom configuration.

This allows us to override the timestamp (this is how posthog-js does it) and nicely allows us to extend the call signatures without just adding a silly number of arguments.

@neilkakkar I had to move out the `forceSendFeatureFlags` code to the posthog-node part only as it feels very specific to local evaluation.
Also I'm not sure if it works... It triggers the `reloadFeatureFlagsAsync` but doesn't use the result it returns? Instead it uses the `getFeatureFlags` code path but that shouldn't be used by node lib as it might have race conditions between different clients. Interested to see your take on this and maybe we need a sync chat about it to figure out the best approach.


posthog-node/src/posthog-node.ts
```diff
@@ -107,12 +107,23 @@ export class PostHog implements PostHogNodeV1 {
...
 
-  capture({ distinctId, event, properties, groups, sendFeatureFlags }: EventMessageV1): void {
+  capture({ distinctId, event, properties, groups, sendFeatureFlags, timestamp }: EventMessageV1): void {
     this.reInit(distinctId)
     if (groups) {
       this._sharedClient.groups(groups)
     }
-    this._sharedClient.capture(event, properties, sendFeatureFlags || false)
+
+    const _capture = (): void => {
```
neilkakkar, line 116: moving this here works.

Talk to me a bit about the race condition issues?

Is it related to other calls to `reInit` changing the flags because the distinctID changes? Since the `capture` call is not blocking?

Hmm, that makes sense

benjackwhite: Yeah exactly. It's one of the things I never really nailed with this combined JS approach 😢 

I think the code will work still if we just change it to use the returned value of the async call?

neilkakkar: Hmm, okay other flag calls are blocking, so this isn't an issue.

PR #45, opened 2022-12-13
Create label-version-bump.yml

timgl:
Direct copy + pasta from https://github.com/PostHog/posthog-js/blob/master/.github/workflows/label-version-bump.yml

### DIFF
--- a/.github/workflows/label-version-bump.yml
+++ b/.github/workflows/label-version-bump.yml
@@ -1,0 +1,75 @@
+name: Autobump
+
+on:
+    pull_request:
+        types: [closed]
+
+jobs:
+    label-version-bump:
+        name: Bump version based on PR label
+        runs-on: ubuntu-20.04
+        if: |
+            github.event.pull_request.merged
+            && (
+                contains(github.event.pull_request.labels.*.name, 'bump patch')
+                || contains(github.event.pull_request.labels.*.name, 'bump minor')
+                || contains(github.event.pull_request.labels.*.name, 'bump major')
+            )
+        steps:
+            - name: Check out repository
+              uses: actions/checkout@v2
+              with:
+                  ref: ${{ github.event.pull_request.base.ref }}
+                  token: ${{ secrets.POSTHOG_BOT_GITHUB_TOKEN }}
+                  fetch-depth: 0
+
+            - name: Detect version bump type
+              id: bump-type
+              run: |
+                  BUMP_TYPE=null
+                  if [[ $BUMP_PATCH_PRESENT == 'true' ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/label-version-bump.yml
```diff
@@ -0,0 +1,75 @@
...
+                  mv package.json package.old.json
+                  jq --indent 4 '.version = "${{ steps.versions.outputs.new-version }}"' package.old.json > package.json
+                  rm package.old.json
+            - name: Update CHANGELOG.md
+              run: |
+                  CHANGELOG_HEADING='## ${{ steps.versions.outputs.new-version }} - '$(date --iso-8601)
+                  CHANGELOG_POINTS=$(git log v${{ steps.versions.outputs.old-version }}..${{ github.event.pull_request.base.ref }} --pretty=format:%s --grep='^.*\d*)$' | sed -e 's/^/- /')
+                  mv CHANGELOG.md CHANGELOG.old.md
+                  echo -e "$CHANGELOG_HEADING\n\n$CHANGELOG_POINTS\n\n$(cat CHANGELOG.old.md)" > CHANGELOG.md
+                  rm CHANGELOG.old.md
```
Twixes, line 65: We probably don't want this step because this repo has no `CHANGELOG.md`

Twixes (commented):
This will work (except the `CHANGELOG.md` part), but `label-version-bump.yml` is only the "Create bump commit on base branch" part of the equation. Don't we also want the https://github.com/PostHog/posthog-js/blob/master/.github/workflows/cd.yml part?

neilkakkar:
Flyby: This repo publishes 3 different packages with 3 different versions, so manually updating `package.json` and letting CI take care of publishing has been more reliable / useful.

.github/workflows/label-version-bump.yml
neilkakkar: (it's in the relevant folder for the relevant package)

Twixes:
Ah, so this is like a monorepo. 🤔 But versions aren't synced between the individual packages and the root. So I'm not sure how something like `label-version-bump.yml` should work. The packages have a shared core, so they should probably all have the same latest version at a time. That means this workflow should update _each_ `package.json`.
 
Also I see there already is a `release.yml` doing the same thing as `cd.yml`, so that point is off the table.

benjackwhite (changes requested):
I like the effort here but not sure if it fits for the repo style we have.

We already have a release flow that detects when the package versions are changed. We _could_ look at doing some additional automation that reacts to changes and sets up the relevant PRs for the affected repos but I think that's fairly different to this.

PR #47, opened 2023-01-04
Respect autocapture=false

stefanmaric:
The `autocapture` prop is typed as `boolean | PostHogAutocaptureOptions`, but passing `false` does nothing at all.

This PR makes the provider behave more intuitively when autocapture is false.

If ignoring the `false` option was intended, the `autocapture` prop should have been typed as `true | PostHogAutocaptureOptions`.


benjackwhite:
I think this makes sense 👍 Seems like we never accounted for someone specifically setting false. If you can fix the linting, I can approve and merge, otherwise I will fork this and try and add a unit test for that logic so that we catch this better in the future.

stefanmaric:
@benjackwhite I fixed formatting. Owe you the tests tho. :smile:

benjackwhite (approved):

PR #50, opened 2023-01-09
feat(feature-flags): v3 feature flag payloads

EDsCODE:
Two new functions based on https://github.com/PostHog/posthog/pull/13580:

`getFeatureFlagPayloads(): PostHogDecideResponse['featureFlagPayloads'] | undefined `  — returns all matched flag payloads
`getFeatureFlagPayload(key:string): JsonType | undefined`  — return only the payload of the flag requested in parameter


posthog-core/src/index.ts
```diff
@@ -478,6 +491,33 @@ export abstract class PostHogCore {
...
+      this.capture('$feature_flag_payload_called', {
+        $feature_flag: key,
+        $feature_flag_payload: response,
+      })
+    }
+
+    return response
+  }
+
+  getFeatureFlagPayloads(): PostHogDecideResponse['featureFlagPayloads'] | undefined {
```
neilkakkar, line 514: curious how you think this function will be used? a.k.a. do we need it? 🤔

posthog-core/src/index.ts
```diff
@@ -478,6 +491,33 @@ export abstract class PostHogCore {
...
+
+    if (!payloads) {
+      return undefined
+    }
+
+    let response = payloads[key]
+
+    if (this.sendFeatureFlagEvent && !this.flagPayloadCallReported[key]) {
+      this.flagPayloadCallReported[key] = true
+      this.capture('$feature_flag_payload_called', {
```
neilkakkar, line 505: thoughts on consuming this into `$feature_flag_called` ? To a user, calling `getFeatureFlag`, or `getFeatureFlagPayload` seems equivalent. We can probs just add the payload parameter to feature_flag_called?

Also ensures no duplicate events sent when we call both `getFeatureFlag` and `getFeatureFlagPayload`

posthog-core/src/index.ts
```diff
@@ -436,6 +438,10 @@ export abstract class PostHogCore {
           this.setKnownFeatureFlags(res.featureFlags)
         }
 
+        if (res.featureFlagPayloads) {
```
neilkakkar, line 441: need to bump decide version too! Also, don't publish new version until my changes go in too 😅

posthog-core/src/index.ts
EDsCODE: Yes, but a little confused what the implementation should look like here. Would `getFeatureFlag ` and `getFeatureFlagPayload ` have the same exact logic now and just return different val since they both need to send the same event

neilkakkar: Hmm tbh, we can maybe get rid of this completely? Don't need to send these extra events 😅 . I kinda like the idea of everything being determined just by the variant, so there shouldn't be a need to audit the payload?

If you do decide to send them, yeah I was thinking the implementation looks the same, with the difference being `getFeatureFlag` returns the `$feature_flag_response` value while `getFeatureFlagPayload` returns the `$feature_flag_payload` 🤷 .

EDsCODE:
Local eval for node suggestion: https://github.com/PostHog/posthog-js-lite/pull/54

posthog-core/src/index.ts
neilkakkar: Alternatively, merge this in after this: https://github.com/PostHog/posthog-js-lite/pull/53/files

EDsCODE:
Waiting on https://github.com/PostHog/posthog-js-lite/pull/53

posthog-core/src/index.ts
EDsCODE: Yup waiting on that one

posthog-core/src/index.ts
EDsCODE: Removed

posthog-core/src/index.ts
```diff
@@ -32,6 +33,7 @@ export abstract class PostHogCore {
   private captureMode: 'form' | 'json'
   private sendFeatureFlagEvent: boolean
   private flagCallReported: { [key: string]: boolean } = {}
+  private flagPayloadCallReported: { [key: string]: boolean } = {}
```
neilkakkar, line 36: rm

posthog-core/src/index.ts
```diff
@@ -443,6 +444,19 @@ export abstract class PostHogCore {
...
+          let newFeatureFlagPayloads = res.featureFlagPayloads
+          if (res.errorsWhileComputingFlags) {
+            // if not all flags were computed, we upsert flags instead of replacing them
+            const currentFlagPayloads = this.getPersistedProperty<PostHogDecideResponse['featureFlagPayloads']>(
+              PostHogPersistedProperty.FeatureFlagPayloads
+            )
+            newFeatureFlagPayloads = { ...currentFlagPayloads, ...res.featureFlagPayloads }
+          }
+          this.setKnownFeatureFlagPayloads(newFeatureFlagPayloads)
+        }
```
EDsCODE, line 458: Do we need upserting here

neilkakkar: Yeah. Actually, can we not duplicate upserting code and move it into the above? wherever flags are set, set the payloads as well, otherwise don't do anything.

Let's always keep flags response as the source of truth, and payloads just follow whatever happens to flags.

neilkakkar: Otherwise, will be easy to make these go out of sync, which is terrible :/

neilkakkar:
Thinking about this further, wherever we set `setKnownFeatureFlags`, we should be setting `setKnownFlagPayloads`, so makes sense to make these private, and the public one takes both.

Extending this thought further, sounds like we need to add payloads to bootstrapping as well, which makes sense 😅 - so a `getAllFlags` companion, `getAllFlagPayloads`, will help populate these.

PR #54, opened 2023-01-10
feat(feature-flags): posthog-node local evaluation

EDsCODE:


posthog-node/src/posthog-node.ts
```diff
@@ -203,6 +204,32 @@ export class PostHog implements PostHogNodeV1 {
     return response
   }
 
+  async getFeatureFlagPayload(key: string, distinctId: string, value: string | boolean, options?: {
```
EDsCODE, line 207: I'm thinking the API looks like this where the user can pass in the matched value and we just handle the lookup. alternatively, we can remove this argument and have this function handle/trigger a compute for the matched flag and then lookup too

posthog-node/src/feature-flags.ts
```diff
@@ -62,6 +63,7 @@ class FeatureFlagsPoller {
     this.pollingInterval = pollingInterval
     this.personalApiKey = personalApiKey
     this.featureFlags = []
+    this.featureFlagPayloads = {}
```
neilkakkar, line 66: quick flyby that the payloads should already be defined in `this.featureFlags` so we shouldn't need this? 🤔

posthog-node/src/feature-flags.ts
```diff
@@ -112,6 +114,35 @@ class FeatureFlagsPoller {
     return response
   }
 
+  async getFeatureFlagPayload(
```
neilkakkar, line 117: for the interface, we'll need a distinct_id here as well, since node is stateless.

Not sure what `value` represents?

posthog-node/src/feature-flags.ts
neilkakkar: oh nvm, you're generating this 👍

posthog-node/src/feature-flags.ts
neilkakkar: ugh, nvm, this isn't the interface 🤦 . I'm going to stay logged off tonight and check this tomorrow 😂

posthog-node/src/posthog-node.ts
neilkakkar: Only for the multivariate case I guess? As otherwise there's only one payload?

I think it makes sense to stick to the same interface as for js, so just key and distinctId. This seems to make it more error prone.

posthog-core/src/index.ts
```diff
@@ -508,6 +508,11 @@ export abstract class PostHogCore {
       })
     }
 
+    // Undefined means a loading or missing data issue. Null means evaluation happened and there was no match
```
EDsCODE, line 511: Not 100% on this

posthog-node/src/feature-flags.ts
EDsCODE: just iterating through them once upon retrieval so it doesn't need to happen everytime

posthog-node/src/posthog-node.ts
```diff
@@ -203,6 +204,69 @@ export class PostHog implements PostHogNodeV1 {
...
+  ): Promise<JsonType | undefined> {
+    const { groups, personProperties, groupProperties } = options || {}
+    let { onlyEvaluateLocally, sendFeatureFlagEvents } = options || {}
+    let response = undefined
+
+    // Try to get match value locally if not provided
+    if (!matchValue) {
+      matchValue = await this.getFeatureFlag(key, distinctId, {
+        ...options,
+        onlyEvaluateLocally: true,
```
neilkakkar, line 227: You can simplify this quite a bit by passing in `onlyEvaluateLocally` here. What you really care about is the `matchValue`, so the burden of local evaluation can be completely passed to `getFeatureFlag`

posthog-core/src/index.ts
neilkakkar: looks like a duplicate?

posthog-node/src/feature-flags.ts
```diff
@@ -112,6 +114,29 @@ class FeatureFlagsPoller {
     return response
   }
 
+  async getFeatureFlagPayload(key: string, matchValue: string | boolean): Promise<JsonType | undefined> {
```
neilkakkar, line 117: open question remains, to pass `matchValue` in or not. I'm still on the side of not passing in, as your argument around "makes for zero user erroneous input" stands strong in server-side libs too.

posthog-node/src/posthog-node.ts
EDsCODE: Hmm, I made it explicitly local on that call because I don't want to defer to decide endpoint as a backup until the very end of the function. So the flow is meant to be -> try everything local first (get value, then get payload) if that's not possible, then we do reinit the shared client which will trigger the decide endpoint call and then return the payload that's desired. If i just forward `onlyEvaluateLocally`, the decide endpoint will be called on getFeatureFlag but the next statement where it tries to retrieve payload locally could fail and then the decide would get hit again upon shared client logic

RE: `should do getFeatureFlagPayloads without matchValue` this test should show the problem. We want to limit it to 1 call and not accidentally hit 2

posthog-node/src/feature-flags.ts
EDsCODE: This line you're referring to is the poller so it requires matchValue but in terms of the interface, is it detrimental to leave the implementation as is? We can document the usage as if you don't pass it in, but also add documentation if it seems it's a legit use case

posthog-node/src/posthog-node.ts
neilkakkar: Not sure I understand.

If matchValue isn't available locally, you'll have to go anyway to decide, yes? And if it's locally evaluable, you wouldn't go to decide anyway.

> If i just forward onlyEvaluateLocally, the decide endpoint will be called on getFeatureFlag but the next statement where it tries to retrieve payload locally could fail and then the decide would get hit again upon shared client logic

Ah, this seems like a bug if it happens? Why would that next statement to retrieve payload locally fail, once we just went to decide?

posthog-node/src/feature-flags.ts
neilkakkar: Ah, this isn't exposed to the client 🤦 - yeah it's perfectly fine then!

posthog-node/src/posthog-node.ts
EDsCODE: Different explanation: if decide needs to happen, we only want it to happen once so it may as well be the right call that will return the payload corresponding to the flag key. I'm intending for `getFeatureFlag` _and_ `getFeatureFlagPayload` to be the logic that tries to handle verything locally, if none of that flies, then rollover to the decide based call. So, that's why the first `getFeatureFlag` call is only needed for the local evaluation code path. If local is a bust, we're going to make a call anyway so we may as well let the right payload logic call happen and not multiple decides

neilkakkar: I guess the part I don't still understand is, why do 2 decide calls happen?

Whenever we call `reloadFeatureFlagsAsync`, not only do we update flags, but also all payloads, no matter where it's called from.

neilkakkar (approved):

posthog-core/src/index.ts
```diff
@@ -33,6 +33,7 @@ export abstract class PostHogCore {
   private captureMode: 'form' | 'json'
   private sendFeatureFlagEvent: boolean
   private flagCallReported: { [key: string]: boolean } = {}
+  private flagPayloadCallReported: { [key: string]: boolean } = {}
```
neilkakkar, line 36: this is still around 👀

posthog-core/src/index.ts
neilkakkar: this is repeated code 😅

posthog-node/src/posthog-node.ts
EDsCODE: Summary of call:
The way that poller and decide results are handled are separate. The decide result can be handled better but that's a matter of refactoring that is out of scope for this PR

PR #57, opened 2023-01-16
fix: support for SSR

paolodamico:
This PR refactors the code `posthog-web` so the library can be imported in Next.js projects with SSR. Instead of window having to be defined just from importing the library, this relies on `window` being defined on the constructor (i.e. when an object is initialized).

**Before in a Next.JS project**
```js
import PostHog from 'posthog-js-lite' // <- this would generate an exception because window is not defined
```

**After (this works)**
```js
import PostHog from 'posthog-js-lite';

if (typeof window !== 'undefined') {
    posthog = new PostHog(
        'phc_my_key',
    )
}
```

CC @m1guelpf

### DIFF
--- a/posthog-web/src/posthog-web.ts
+++ b/posthog-web/src/posthog-web.ts
@@ -56,7 +56,7 @@ export class PostHog extends PostHogCore {
   getLibraryVersion(): string {
     return version
   }
-  
+
   getCustomUserAgent(): void {
     return
   }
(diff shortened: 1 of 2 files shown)


posthog-web/src/posthog-web.ts
```diff
@@ -19,7 +19,7 @@ export class PostHog extends PostHogCore {
 
     // posthog-js stores options in one object on
     this._storageKey = options?.persistence_name ? `ph_${options.persistence_name}` : `ph_${apiKey}_posthog`
-    this._storage = getStorage(options?.persistence || 'localStorage')
+    this._storage = getStorage(options?.persistence || 'localStorage', window)
```
paularmstrong, line 22: This still needs to be gated because it is still throwing a ReferenceError – you can't just reference `window` in the node.js global scope. 

```
this._storage = getStorage(options?.persistence || 'localStorage', typeof window !== 'undefined' ? window : undefined)
```

PR #59, opened 2023-01-19
feat(feature-flag): ff bootstrap payload option

EDsCODE:
- Add ability to bootstrap payloads
- init can now accept `featureFlagPayloads: Record<string, JsonType>`

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -550,13 +550,16 @@ export abstract class PostHogCore {
     return flags
   }
 
-  getFeatureFlagsAndPayloads(): {flags: PostHogDecideResponse['featureFlags'] | undefined, payloads: PostHogDecideResponse['featureFlagPayloads'] | undefined} {
+  getFeatureFlagsAndPayloads(): {
+    flags: PostHogDecideResponse['featureFlags'] | undefined
+    payloads: PostHogDecideResponse['featureFlagPayloads'] | undefined
+  } {
     const flags = this.getFeatureFlags()
     const payloads = this.getFeatureFlagPayloads()
 
     return {
       flags,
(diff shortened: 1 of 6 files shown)


posthog-core/src/index.ts
```diff
@@ -116,6 +116,7 @@ export abstract class PostHogCore {
           {}
         )
       this.setKnownFeatureFlags(activeFlags)
+      options?.bootstrap.featureFlagPayloads && this.setKnownFeatureFlagPayloads(options?.bootstrap.featureFlagPayloads)
```
neilkakkar, line 119: we also need a way to _get_ all payloads, alongside `getAllFlags` in server-side libs.

Think it's worth combining those two together, keeps things simple, `getAllFlagsAndPayloads()`, which can then be directly passed to bootstrap.

posthog-node/src/posthog-node.ts
```diff
@@ -338,6 +338,59 @@ export class PostHog implements PostHogNodeV1 {
     return response
   }
 
+  async getAllPayloads(
```
neilkakkar, line 341: Thinking about this a bit more, for the final interface as well, I don't think anyone would want _just_ payloads, so we should turn this into `getAllFlagsAndPayloads` as well. Thoughts?

Imagine someone bootstrapping, ideally they'll just want one request max to get everything they need to pass to the frontend, as this will be very time sensitive.

neilkakkar: and then we can combine the above and this to use `getAllflagsAndPayloads` only, while `getAllFlags` just returns the flags from `getAllflagsAndPayloads`

neilkakkar (approved):

PR #60, opened 2023-01-19
feat(rn): Adds support for Wix's react-native-navigation

benjackwhite:
And an example of how it should be used

Fixes some other things:

## Changelog
* Autocapture now tracks the text content of a pressed view, mimicking what Web autocapture does.
* Adds support for passing a promise to the PostHogProvider to help those who choose to instantiate the PostHog client directly
* Adds support for Wix's react-native-navigation through `initReactNativeNavigation`. Please check the docs for setup guides as there are some specifics required.
* Allows setting `debug` from the PostHogProvider itself.


posthog-react-native/src/PostHogProvider.tsx
```diff
@@ -51,6 +48,31 @@ export const PostHogProvider = ({
...
+        setPosthog(client)
+      }
+    }
+  }, [client])
+
+  useEffect(() => {
+    if (apiKey && !posthog) {
+      PostHog.initAsync(apiKey, options).then(setPosthog)
+    }
+  }, [apiKey])
```
thmsobrmlr, line 66: I think there is a race condition between initializing the client from a given apiKey and an async initialization. Generally I think one would pass either an apiKey or a client, right? Would be good to add a warning for accidentally passing both.

posthog-react-native/package.json
```diff
@@ -34,7 +35,8 @@
     "expo-device": ">= 4.0.0",
     "expo-file-system": ">= 13.0.0",
     "expo-localization": ">= 11.0.0",
-    "react-native-device-info": ">= 10.0.0"
+    "react-native-device-info": ">= 10.0.0",
+    "react-native-navigation": ">=6.0.0"
```
thmsobrmlr, line 39: This should be added as optional to `peerDependenciesMeta` as well, so that users without `react-native-navigation` don't have to add it.

thmsobrmlr (approved):
Looks great! I've left two comments inline, and besides that I'm assuming you have tested that the optional loading of `react-native-navigation` works and it's not accidentally included in the bundle for users who aren't using the library?

posthog-react-native/src/PostHogProvider.tsx
benjackwhite: That's a good point! I will do exactly that

posthog-react-native/package.json
benjackwhite: Ah yep

PR #69, opened 2023-02-03
feat: Allow customising appProperties with a function

benjackwhite:
## Problem

Customer request was basically asking to omit one of the properties we bundle by default. This is currently already possible but only by extending the `PostHog` class and overriding the property like this:

```tsx
class CustomPostHog extends PostHog {
  getCommonEventProperties(): any {
    const properties = super.getCommonEventProperties()
    // Filter out or add more properties here

    delete properties.$device_name

    return properties
  }
}
```

We do provide an option which was intended for providing your own "native" app properties if you don't want to use the libraries we support via [native-deps](https://github.com/PostHog/posthog-js-lite/blob/master/posthog-react-native/src/native-deps.tsx#L9) but this becomes a full override, meaning you have to build all the options yourself manually.

```tsx
PostHog.initAsync("token", { customAppProperties: {
  $device_name: 'My own thing',
}})
```

## Changes

Allow customAppProperties to be a function that takes the existing properties

```tsx

PostHog.initAsync('token', {
  customAppProperties: (props) => ({
    ...props,
    $device_name: 'CUSTOM!',
  }),
})
```

which feels a lot more client-friendly that overriding the class.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- Added support for customising the default app properties by passing a function to `options.customAppProperties`


pauldambra (approved):
🚢

neilkakkar:
Lovely, elegant, I like it!

PR #74, opened 2023-02-13
feat: Make PostHog Core stateless

neilkakkar:
## Problem

This is part 1 of 2 of making posthog core stateless.

The gameplan:

Part 1: Split out two abstract base classes: PostHogCore (existing class), and PostHogCoreStateless (new class, which is also the base class of PosthogCore).

The reasoning behind this is to keep changes minimal, and to ensure that current tests pass without any changes. The PostHogCore interface remains the same, so technically nothing should change because of part 1.

Part 2: In a future PR, I will refactor posthog-node to depend on `PostHogCoreStateless` instead of `PostHogCore`.

### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -24,7 +24,7 @@ export * as utils from './utils'
 import { LZString } from './lz-string'
 import { SimpleEventEmitter } from './eventemitter'
 
-export abstract class PostHogCore {
+export abstract class PostHogCoreStateless {
   // options
   private apiKey: string
   host: string
@@ -32,16 +32,13 @@ export abstract class PostHogCore {
   private flushInterval: number
   private requestTimeout: number
   private captureMode: 'form' | 'json'
-  private sendFeatureFlagEvent: boolean
-  private flagCallReported: { [key: string]: boolean } = {}
   private removeDebugCallback?: () => void
 
   // internal
   protected _events = new SimpleEventEmitter()
   protected _flushTimer?: any
-  protected _decideResponsePromise?: Promise<PostHogDecideResponse>
+  // protected _decideResponsePromise?: Promise<PostHogDecideResponse>
   protected _retryOptions: RetriableOptions
-  protected _sessionExpirationTimeSeconds: number
 
   // Abstract methods to be overridden by implementations
   abstract fetch(url: string, options: PostHogFetchOptions): Promise<PostHogFetchResponse>
@@ -53,7 +50,6 @@ export abstract class PostHogCore {
   abstract getPersistedProperty<T>(key: PostHogPersistedProperty): T | undefined
   abstract setPersistedProperty<T>(key: PostHogPersistedProperty, value: T | null): void
 
-  private _optoutOverride: boolean | undefined
 
   constructor(apiKey: string, options?: PosthogCoreOptions) {
     assert(apiKey, "You must pass your PostHog project's api key.")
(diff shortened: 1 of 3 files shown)


posthog-core/src/index.ts
```diff
@@ -63,38 +57,422 @@ export abstract class PostHogCore {
...
+        event: '$identify',
+        properties,
+      }),
+    }
+
+    this.enqueue('identify', payload, options)
+    return this
+  }
+
+  captureStateless(
```
benjackwhite, line 115: Wondering, should we make these methods `protected`?

When you do the node version, we wouldn't want these methods to be public right? So the only reason to not have them protected is if we want the stateful clients to be able to call the stateless method which I feel like we wouldn't want, or at least the naming wouldn't make sense in that context.

posthog-core/src/index.ts
```diff
@@ -63,38 +57,422 @@ export abstract class PostHogCore {
...
+        ...this.getCommonEventProperties(), // Common PH props
+      },
+    }
+  }
+
+  /***
+   *** TRACKING
+   ***/
+  identifyStateless(distinctId: string, properties?: PostHogEventProperties, options?: PosthogCaptureOptions): this {
+    // This identify doesn't add anything to set, you have to do it yourself!
```
benjackwhite, line 102: I sort of get why this comment is here but I don't fully understand it 😅

posthog-core/src/index.ts
```diff
@@ -63,38 +57,422 @@ export abstract class PostHogCore {
...
+      library_version: this.getLibraryVersion(),
+      timestamp: options?.timestamp ? options?.timestamp : currentISOTime(),
+    }
+
+    if (message.distinctId) {
+      message.distinct_id = message.distinctId
+      delete message.distinctId
+    }
+
+    // TODO: check why this isn't flushed properly....
```
benjackwhite, line 322: Still todo?

posthog-core/src/index.ts
```diff
@@ -63,38 +57,422 @@ export abstract class PostHogCore {
...
+  }
+}
+
+export abstract class PostHogCore extends PostHogCoreStateless {
+  // options
+  private sendFeatureFlagEvent: boolean
+  private flagCallReported: { [key: string]: boolean } = {}
+
+  // internal
+  protected _decideResponsePromise?: Promise<PostHogDecideResponse | undefined> // TODO: come back to this, fix typing
```
benjackwhite, line 458: Still todo?

posthog-core/test/posthog.identify.spec.ts
```diff
@@ -26,15 +26,15 @@ describe('PostHog Core', () => {
...
-            },
             properties: {
               $lib: 'posthog-core-tests',
               $lib_version: '2.0.0-alpha',
               foo: 'bar',
               $anon_distinct_id: expect.any(String),
               $session_id: expect.any(String),
+              $set: {
+                foo: 'bar',
+              },
```
benjackwhite, line 37: Was this an existing bug? 😬 🙈

benjackwhite (changes requested):
Some comments and sanity checks but overall looks great. So glad you are doing this work 🙌

posthog-core/test/posthog.identify.spec.ts
neilkakkar: I just moved it down, because ordering has changed 😬

posthog-core/src/index.ts
neilkakkar: the point is this is the purest form of the fn, so node and core would both call this!

See how node will work around this: https://github.com/PostHog/posthog-js-lite/pull/75

posthog-core/src/index.ts
neilkakkar: yes-ish, want to confirm request structure in a proper example with node once it's in place. Can I leave this in for the follow up? 😅

posthog-core/src/index.ts
neilkakkar: this issue: https://github.com/PostHog/posthog/issues/13092 - I have a test I didn't commit (because it will fail :P )

posthog-core/src/index.ts
neilkakkar: Will fix

posthog-core/src/index.ts
benjackwhite: Ah I misundestood the plan here perhaps... I figured the Node library would `extend` the stateless class and then call the methods that way, meaning they can remain `protected` as we wouldn't need two classes?

Maybe I need to have a deeper look at the other PR...

neilkakkar: Could do it like that, tbh. Yeah, makes sense! Will make these protected in that PR though, if I go down that route!

benjackwhite: Made a PR suggestion to validate my thinking - take it or leave it :D https://github.com/PostHog/posthog-js-lite/pull/76

neilkakkar: took it!

benjackwhite (approved):
Approved assuming we are not going to release this and instead just overwrite with https://github.com/PostHog/posthog-js-lite/pull/75

PR #81, opened 2023-02-16
Properties file storage

davidthor:
## Problem

Command line interfaces, and other processes that terminate after each call, don't persist the properties in a way that simulates the idea of a "session". We need a way for CLIs to use the same anonymous ID with each call so we can sanely correlate the ID with a distinct ID at a later time when they run a login command.

## Changes

Introduces the class, `PostHogFsStorage` into posthog-core and updated posthog-node to use the `persistence` option to toggle between in-memory storage and filesystem storage of properties like the anonymous_id.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for filesystem properties persistence for the posthog-node library


posthog-node/src/posthog-node.ts
```diff
@@ -9,13 +9,15 @@ import {
...
+import { PostHogMemoryStorage } from '../../posthog-core/src/storage/storage-memory'
+import { PostHogFsStorage } from '../../posthog-core/src/storage/storage-fs'
 import { EventMessageV1, GroupIdentifyMessage, IdentifyMessageV1, PostHogNodeV1 } from './types'
 import { FeatureFlagsPoller } from './feature-flags'
 import { fetch } from './fetch'
+import { PostHogStorage } from 'posthog-core/src/storage/storage'
 
 export type PostHogOptions = PosthogCoreOptions & {
-  persistence?: 'memory'
+  persistence?: string
```
neilkakkar, line 20: Let's keep this a literal with possible options, and add another configuration option to pass in the file path

neilkakkar (changes requested):
This is great, thanks for the contribution! 🚀 

Would love to see some tests, like here: https://github.com/PostHog/posthog-js-lite/blob/master/posthog-react-native/test/storage.spec.ts#L4

posthog-node/src/posthog-node.ts
davidthor: How's the new interface look?

davidthor:
> This is great, thanks for the contribution! 🚀
> 
> Would love to see some tests, like here: https://github.com/PostHog/posthog-js-lite/blob/master/posthog-react-native/test/storage.spec.ts#L4

It doesn't look like there's currently a basis for testing storage/persistence in the node package (looks like you linked to the react-native package which this doesn't touch). Any suggestions or best practices for testing would be most appreciated.

neilkakkar:
Yeah, I meant the exact same file, but for posthog-node (and FsStorage). Things should be similar enough, thanks to similar interfaces, that the file should be a good indicator of what to do for node :)

posthog-node/src/posthog-node.ts
neilkakkar: looks good!

davidthor:
I added a test suite for the file storage class, but I can't seem to run the full test suite locally. I get the following error on 17 of the different test files:

```
TypeError: Cannot assign to read only property 'performance' of object '[object global]'

      at hijackMethod (node_modules/@sinonjs/fake-timers/src/fake-timers-src.js:946:32)
      at Object.install (node_modules/@sinonjs/fake-timers/src/fake-timers-src.js:1733:17)
      at FakeTimers.useFakeTimers (node_modules/@jest/fake-timers/build/modernFakeTimers.js:110:36)
```

benjackwhite (changes requested):

PR #82, opened 2023-02-17
fix: Identify $set for Node

benjackwhite:
## Problem

Stateless refactor introduced a bug with identify on node

## Changes

* Fix this to use $set and also handle if someone treats it like a `capture` call.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### DIFF
--- a/posthog-node/test/posthog-node.spec.ts
+++ b/posthog-node/test/posthog-node.spec.ts
@@ -114,7 +114,27 @@ describe('PostHog Node.js', () => {
           distinct_id: '123',
           event: '$identify',
           properties: {
-            foo: 'bar',
+            $set: {
+              foo: 'bar',
+            },
+          },
+        },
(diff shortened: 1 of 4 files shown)


neilkakkar (approved):
Perfect, thank you!

posthog-node/test/posthog-node.spec.ts
```diff
@@ -114,7 +114,27 @@ describe('PostHog Node.js', () => {
           distinct_id: '123',
           event: '$identify',
           properties: {
-            foo: 'bar',
+            $set: {
```
neilkakkar, line 117: oh, looks like there already _was_ a test here and I messed up 🤔

PR #91, opened 2023-04-20
fix: Race conditions when initialising the RN SDK

benjackwhite:
## Problem

Customer noticed that there seemed to be some ID issues despite using `PostHog.initAsync`. Investigated and found that the issue is happening as we don't wait for the async storage to be loaded _before_ we init PostHog and in turn call the reload.

## Changes

* Changes `.initAsync` to warn if the SDK has already been initialised
* Moves the preload await to _before_ instantiating PostHog which ensures the correct IDs are loaded
* Handles the preloading of FF in the RN constructor to ensure we definitely don't accidentally load too early

- [x] TODO: Fix the reset logic so that we reload flags on subsequent identify

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [x] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- Fixes a race condition that could occur when initialising PostHog
- Fixes an issue where feature flags would not be reloaded after a reset
- (RN): PostHog should always be initialized via `.initAsync` and will now warn if this is not the case.

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -587,10 +587,6 @@ export abstract class PostHogCore extends PostHogCoreStateless {
 
     this.sendFeatureFlagEvent = options?.sendFeatureFlagEvent ?? true
     this._sessionExpirationTimeSeconds = options?.sessionExpirationTimeSeconds ?? 1800 // 30 minutes
-
-    if (options?.preloadFeatureFlags !== false) {
-      this.reloadFeatureFlags()
-    }
   }
 
   protected setupBootstrap(options?: Partial<PosthogCoreOptions>): void {
@@ -740,9 +736,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
       this.setPersistedProperty(PostHogPersistedProperty.AnonymousId, previousDistinctId)
       this.setPersistedProperty(PostHogPersistedProperty.DistinctId, distinctId)
 
-      if (this.getFeatureFlags()) {
-        this.reloadFeatureFlags()
-      }
+      this.reloadFeatureFlags()
     }
 
     super.identifyStateless(distinctId, allProperties, options)
@@ -809,7 +803,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
       },
     })
 
-    if (Object.keys(groups).find((type) => existingGroups[type] !== groups[type]) && this.getFeatureFlags()) {
+    if (Object.keys(groups).find((type) => existingGroups[type] !== groups[type])) {
       this.reloadFeatureFlags()
     }
 
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -556,7 +556,13 @@ export abstract class PostHogCoreStateless {
     clearTimeout(this._flushTimer)
     try {
       await this.flushAsync()
-      await Promise.all(Object.values(this.pendingPromises))
+      await Promise.all(
+        Object.values(this.pendingPromises).map((x) =>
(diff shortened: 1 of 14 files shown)


neilkakkar (approved):
Makes sense, nice work!

posthog-react-native/src/posthog-rn.ts
```diff
@@ -57,27 +75,26 @@ export class PostHog extends PostHogCore {
...
-    } else {
-      void this._semiAsyncStorage.preloadAsync().then(() => {
-        setupFromStorage()
-      })
+      if (options?.preloadFeatureFlags !== false) {
+        this.reloadFeatureFlags()
+      }
     }
+
+    void setupAsync()
```
neilkakkar, line 97: How do people know when things are in 'ready' state with react native?

I expect everything to bork till this hasn't fulfilled?

posthog-react-native/src/storage.ts
```diff
@@ -17,26 +17,35 @@ export class SemiAsyncStorage {
...
         }
-
-        this.isPreloaded = true
       } catch (e) {
         console.warn(
           "PostHog failed to load persisted data from storage. This is likely because the storage format is. We'll reset the storage.",
           e
         )
+      } finally {
+        this.isPreloaded = true
```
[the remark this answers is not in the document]
benjackwhite: It's a bit unobvious but basically I realised even if parsing fails we should call it "preloaded" because there is no other state that can be handled other than resetting the state

posthog-react-native/src/posthog-rn.ts
benjackwhite: Indeed. I added a warning above this block to basically say "hey - it looks like you didn't use initAsync (which you should"

posthog-core/src/index.ts
```diff
@@ -556,7 +556,13 @@ export abstract class PostHogCoreStateless {
     clearTimeout(this._flushTimer)
     try {
       await this.flushAsync()
-      await Promise.allSettled(Object.values(this.pendingPromises))
```
benjackwhite, line 559: allSettled is not always available so I changed to all with a catch handler

posthog-core/src/index.ts
```diff
@@ -587,13 +593,6 @@ export abstract class PostHogCore extends PostHogCoreStateless {
 
     this.sendFeatureFlagEvent = options?.sendFeatureFlagEvent ?? true
     this._sessionExpirationTimeSeconds = options?.sessionExpirationTimeSeconds ?? 1800 // 30 minutes
-
-    // NOTE: It is important we don't initiate anything in the constructor as some async IO may still be underway on the parent
-    if (options?.preloadFeatureFlags !== false) {
-      safeSetTimeout(() => {
-        this.reloadFeatureFlags()
-      }, 1)
-    }
```
benjackwhite, line 596: This is the source of all our problems so moved this to be something the individual libs deal with

posthog-core/test/posthog.featureflags.spec.ts
```diff
@@ -16,6 +16,7 @@ describe('PostHog Core', () => {
     'json-payload': true,
   })
 
+  // NOTE: Aren't these supposed to be strings?
```
benjackwhite, line 19: Comment for the FF team that I pinged them about - our tests are pre-parsed when they should be json strings

neilkakkar (approved):
will leave the payload fixes for a separate PR, let's get this in first!

PR #100, opened 2023-05-25
feat: Save the $screen_name on every event

benjackwhite:
## Problem

Customer brought up that autocapture events would be way more useful with the current screen name available.

## Changes

* Registers the $screen_name whenever screen is called for subsequent events

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- The `$screen_name` property will be registered for all events whenever `screen` is called

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -619,7 +619,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
   }
 
   // NOTE: Props are lazy loaded from localstorage hence the complex getter setter logic
-  private get props(): PostHogEventProperties {
+  public get props(): PostHogEventProperties {
     if (!this._props) {
       this._props = this.getPersistedProperty<PostHogEventProperties>(PostHogPersistedProperty.Props)
     }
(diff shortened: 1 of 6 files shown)


posthog-core/src/index.ts
```diff
@@ -707,12 +707,15 @@ export abstract class PostHogCore extends PostHogCoreStateless {
     return this.getPersistedProperty<string>(PostHogPersistedProperty.DistinctId) || this.getAnonymousId()
   }
 
-  register(properties: { [key: string]: any }): void {
+  register(properties: { [key: string]: any }, options?: { persist: boolean }): void {
```
neilkakkar, line 710: Maybe I'm missing something but if I do:

1. `register({non_persisted_property: 'xyz'}, {persist: false})`
2. `register({persisted_property: 'xyz'})`

Then it will persist the non-persisted property too?

benjackwhite: Ooooohhh good catch! Yep. Looks like we need a separate non-persisted props object...

neilkakkar (approved):

Issue #107, opened 2023-08-25
posthog-react-native 2.X doesn't send autocaptured “Application Installed” event


robbie-c:
## Bug description

The native SDKs send this
* ios: https://github.com/PostHog/posthog-ios/blob/3a6afc24d6bde730a19470d4e6b713f44d076ad9/PostHog/Classes/PHGPostHog.m#L140
* android: https://github.com/PostHog/posthog-android/blob/09940e6921bafa9e01e7d68b8c9032671a21ae73/posthog/src/main/java/com/posthog/android/PostHog.java#L278

Which means that 1.X of posthog-react-native (which was a wrapper around the iOS/Android SDK) sent it too. 2.X does not send this, and some users who have upgraded to 2.X have asked about this:
* https://posthoghelp.zendesk.com/agent/tickets/5304
* https://posthog.slack.com/archives/C04BBHRSW5A/p1691393804386489

## How to reproduce

1. Add posthog-react-native@1.1.5 to app
2. Install app
3. Observe events in dashboard
4. Add posthog-react-native@2.7.1 to app
2. Install app
3. Observe (lack of) events in dashboard

## Related sub-libraries

- [x] posthog-react-native

PR #108, opened 2023-08-25
feat: Add react native autocapture lifecycle events

robbie-c:
## Problem
Fixes https://github.com/PostHog/posthog-js-lite/issues/107

~Note: please merge https://github.com/PostHog/posthog-js-lite/pull/109 first, as this depends upon that.~
## Changes

Adds these events that were present in v1. I put it behind an options flag, so that users will need to opt-in. This will prevent people from getting an unexpected surge of these events when updating the sdk.

Note: I found that due to ignoring the returned promise on line https://github.com/PostHog/posthog-js-lite/blob/54178fc167362f550c7f5703f2931abc487f8134/posthog-react-native/src/posthog-rn.ts#L108 , `initAsync` would return before `setupAsync` had completed. This is something that is a bit awkward to get right, when there in a constructor which needs to do something async. I fixed this by storing the promise and awaiting it in `initAsync`.

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- Added an options flag `captureNativeAppLifecycleEvents` which, when set, will cause the SDK to automatically capture 'Application Installed', 'Application Updated', 'Application Opened', and 'Application Backgrounded'.

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -93,6 +93,10 @@ export abstract class PostHogCoreStateless {
     }
     this.requestTimeout = options?.requestTimeout ?? 10000 // 10 seconds
     this.disableGeoip = options?.disableGeoip ?? true
+
+    if (options?.__onConstructed) {
+      options.__onConstructed(this)
+    }
   }
 
   protected getCommonEventProperties(): any {
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -93,10 +93,6 @@ export abstract class PostHogCoreStateless {
     }
     this.requestTimeout = options?.requestTimeout ?? 10000 // 10 seconds
     this.disableGeoip = options?.disableGeoip ?? true
-
-    if (options?.__onConstructed) {
-      options.__onConstructed(this)
-    }
   }
 
   protected getCommonEventProperties(): any {
--- a/posthog-react-native/src/posthog-rn.ts
+++ b/posthog-react-native/src/posthog-rn.ts
@@ -1,4 +1,4 @@
-import { AppState, Dimensions } from 'react-native'
+import { AppState, Dimensions, Linking } from 'react-native'
 
 import {
   PostHogCore,
@@ -106,7 +106,7 @@ export class PostHog extends PostHogCore {
       }
 
       if (options?.autoCaptureInstallationEvents) {
-        this.autoCaptureInstallationEvents()
+        this.autoCaptureNativeAppLifecycleEvents()
       }
     }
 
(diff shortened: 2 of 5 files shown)


posthog-react-native/src/posthog-rn.ts
```diff
@@ -55,7 +57,9 @@ export class PostHog extends PostHogCore {
       console.warn('PostHog.initAsync called twice with the same apiKey. The first instance will be used.')
     }
 
-    return posthog
+    const resolved = await posthog
+    await resolved._setupPromise
+    return resolved
```
robbie-c, line 62: Here's the async fix

posthog-react-native/src/posthog-rn.ts
```diff
@@ -160,4 +168,65 @@ export class PostHog extends PostHogCore {
...
+          version: appVersion,
+          build: appBuild,
+          from_background: true,
+        })
+      } else if (state === 'background') {
+        this.capture('Application Backgrounded')
+      }
+    })
+
+    this.setPersistedProperty(PostHogPersistedProperty.InstalledAppBuild, appBuild)
```
benjackwhite, line 229: Given that this logic is super dependent on persisted values, maybe we need a check to make sure that we have "real" storage?

posthog-react-native/src/posthog-rn.ts
```diff
@@ -22,6 +22,7 @@ export type PostHogOptions = PosthogCoreOptions & {
     | PostHogCustomAppProperties
     | ((properties: PostHogCustomAppProperties) => PostHogCustomAppProperties)
   customAsyncStorage?: PostHogCustomAsyncStorage
+  captureNativeAppLifecycleEvents?: boolean
```
benjackwhite, line 25: I feel like the naming here clashes with the PostHogAutocaptureOptions. No good option comes to mind but it feels like we might get messier and messier over time... I wonder if we shouldn't just include them as part of the "lifecycle" events, even though they're not strictly dependent on react lifecycles...

posthog-core/src/index.ts
```diff
@@ -93,6 +93,10 @@ export abstract class PostHogCoreStateless {
     }
     this.requestTimeout = options?.requestTimeout ?? 10000 // 10 seconds
     this.disableGeoip = options?.disableGeoip ?? true
+
+    if (options?.__onConstructed) {
+      options.__onConstructed(this)
+    }
```
benjackwhite, line 99: This feels like a hack and I'm not sure what it is providing different from awaiting the SDK via initAsync? At some point someone will see this and use it, even though its meant to be "just for testing". 
Is there no alternative?

benjackwhite (changes requested):

posthog-react-native/src/posthog-rn.ts
```diff
@@ -160,4 +168,65 @@ export class PostHog extends PostHogCore {
...
+      console.warn(
+        'PostHog could not track installation/update/open, as the build and version were not set. ' +
+          'This can happen if some dependencies are not installed correctly, or if you have provided' +
+          'customAppProperties but not included $app_build or $app_version.'
+      )
+    }
+    if (appBuild) {
+      if (!prevAppBuild) {
+        // new app install
+        this.capture('Application Installed', {
```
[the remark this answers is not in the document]
robbie-c: This function is only called when people explicitly set `options?.captureNativeAppLifecycleEvents`, so anyone doing an update without reading the docs and making code changes just won't get these events.

Maybe one slight optimization would be storing the build number in storage regardless, so that people enabling this flag in future don't get a flood of events.

posthog-react-native/src/posthog-rn.ts
robbie-c: I think we know this just based on it being a React Native app, we use `expo-file-system` or `@react-native-async-storage/async-storage` and throw if neither one is available. The user can provide their own but if they do that I don't think it's unreasonable to epxect that it's "real"

robbie-c: Warning added

posthog-react-native/src/posthog-rn.ts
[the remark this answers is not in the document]
robbie-c: Persistance now happens regardless

posthog-core/src/index.ts
robbie-c: It's now gone, I do call a private method in test code now though

posthog-react-native/src/posthog-rn.ts
robbie-c: I wanted it to be a separate flag so as to not spam people who update. We could definitely not do that and have it under the same flag, but as you mentioned above we would need to be careful about how we communicate the change, whereas with a new flag it's less of a big deal.

benjackwhite (approved):
Looks good! Thanks for the back and forth here

Issue #121, opened 2023-10-27
[posthog-node] Upgrade axios to 1.6.0


arvidboivie:
## Bug description

axios 1.6.0 includes a fix for https://security.snyk.io/vuln/SNYK-JS-AXIOS-6032459

posthog-node is currently on `^0.27.0`

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

benjackwhite:
Not sure if that vulnerability really affects us but happy to do the upgrade

PR #122, opened 2023-10-27
chore: Upgrade axios dep

benjackwhite:
## Problem

Fixes https://github.com/PostHog/posthog-js-lite/issues/121

## Changes

User mentioned our axios dep is somewhat outdated. The vulnerability mentioned shouldn't affect us given we only call our own endpoint, but nonetheless worth keeping up to date.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### Testing

Tested locally with the example and all seemed fine

### DIFF
--- a/examples/example-node/yarn.lock
+++ b/examples/example-node/yarn.lock
@@ -2,6 +2,11 @@
 # yarn lockfile v1
 
 
+"@aashutoshrathi/word-wrap@^1.2.3":
+  version "1.2.6"
+  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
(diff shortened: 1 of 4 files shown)


pauldambra (approved):
🤷

huw:
Ack! This breaks all non-Node runtimes (ex. Cloudflare) because [Axios v1 doesn’t provide the correct non-Node ESM exports](https://github.com/axios/axios/issues/5495), which leads them to try and import a bunch of node modules, which they can’t.

A few potential fixes off the top of my head:
1. Replace Axios with `fetch` (now that `fetch` is in all LTS versions, this might not be unreasonable; it would also mean non-Node runtimes don’t have to add `fetch` as a parameter when instantiating PostHog)
2. Agitate Axios to fix axios/axios#5495
3. Patch it
4. Split the package into Node and non-Node in `package.json#exports`

PR #141, opened 2023-12-18
feat: Sentry integration

### LINKED ISSUE #106: Sentry Integration for `posthog-node`

benjackwhite:
Clone of https://github.com/PostHog/posthog-js-lite/pull/120

## Changes

Resolves #106 by porting the Sentry integration from `posthog-js` to `posthog-js-lite`. I have decided to port this for Node only for now, since I don’t think (??) `posthog-js` is in production yet. If we were to rely on a common core here some of the code would have to change.

I am also unsure of how this library wants to differ from `posthog-js`—the main edits I made were stylistic, and I tried to type everything a bit more strongly, which was the… vibe I’m getting from this repo? So I’m submitting this PR early and we can figure out what you actually want here (I also didn’t write tests). I will leave comments with questions.

## Testing Strategy
I tested this manually by compiling the library & patching it into my own app. I tested that:

- Sentry URL is attached to PostHog events and resolves correctly (only on `sentry.io`)
- PostHog Person URL is attached to Sentry events and resolves correctly (only on `eu.posthog.com`)

I am happy to write a test (how?) or test more extensively, wanted to wait until we had a more-reviewed PR before we do that.

(@benjackwhite - added tests)

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added `posthog.SentryIntegration` to `posthog-node`. Usage:

```ts
Sentry.init({
     dsn: 'https://example',
     integrations: [
         new posthog.SentryIntegration(posthog)
     ]
})

Sentry.setTag(posthog.POSTHOG_ID_TAG, 'some distinct id');
```

### DIFF
--- a/examples/example-node/server.ts
+++ b/examples/example-node/server.ts
@@ -27,10 +27,6 @@ posthog.debug()
 
 Sentry.init({
   dsn: 'https://examplePublicKey@o0.ingest.sentry.io/0',
-
-  // We recommend adjusting this value in production, or using tracesSampler
-  // for finer control
-  tracesSampleRate: 1.0,
   integrations: [new PostHogSentryIntegration(posthog)],
 })
 
--- a/examples/example-node/server.ts
+++ b/examples/example-node/server.ts
@@ -1,6 +1,8 @@
 import express from 'express'
-import { PostHog } from 'posthog-node'
+import { PostHog, PostHogSentryIntegration } from 'posthog-node'
 import undici from 'undici'
+
+import * as Sentry from '@sentry/node'
 
 const app = express()
 
@@ -23,9 +25,27 @@ const posthog = new PostHog(PH_API_KEY, {
 
 posthog.debug()
 
+Sentry.init({
+  dsn: 'https://examplePublicKey@o0.ingest.sentry.io/0',
+
+  // We recommend adjusting this value in production, or using tracesSampler
+  // for finer control
+  tracesSampleRate: 1.0,
+  integrations: [new PostHogSentryIntegration(posthog)],
+})
+
 app.get('/', (req, res) => {
   posthog.capture({ distinctId: 'EXAMPLE_APP_GLOBAL', event: 'legacy capture' })
   res.send({ hello: 'world' })
+})
+
+app.get('/error', (req, res) => {
+  Sentry.captureException(new Error('example error'), {
+    tags: {
+      [PostHogSentryIntegration.POSTHOG_ID_TAG]: 'EXAMPLE_APP_GLOBAL',
+    },
+  })
+  res.send({ status: 'error!!' })
 })
 
 app.get('/user/:userId/action', (req, res) => {
--- a/posthog-node/src/extensions/sentry-integration.ts
+++ b/posthog-node/src/extensions/sentry-integration.ts
@@ -48,11 +48,11 @@ interface PostHogSentryExceptionProperties {
  *     Sentry.init({
  *          dsn: 'https://example',
  *          integrations: [
- *              new posthog.SentryIntegration(posthog)
+ *              new PostHogSentryIntegration(posthog)
  *          ]
  *     })
  *
- *     Sentry.setTag(posthog.POSTHOG_ID_TAG, 'some distinct id');
+ *     Sentry.setTag(PostHogSentryIntegration.POSTHOG_ID_TAG, 'some distinct id');
  *
  * @param {Object} [posthog] The posthog object
  * @param {string} [organization] Optional: The Sentry organization, used to send a direct link from PostHog to Sentry
--- a/posthog-node/src/extensions/sentry-integration.ts
+++ b/posthog-node/src/extensions/sentry-integration.ts
@@ -3,8 +3,6 @@
  */
 import { type PostHog } from '../posthog-node'
 
-export const POSTHOG_ID_TAG = 'posthog_distinct_id'
-
 // NOTE - we can't import from @sentry/types because it changes frequently and causes clashes
 // We only use a small subset of the types, so we can just define the integration overall and use any for the rest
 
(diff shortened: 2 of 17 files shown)


huw:
Easy, I’m following this one now

examples/example-node/server.ts
```diff
@@ -23,11 +25,29 @@ const posthog = new PostHog(PH_API_KEY, {
 
 posthog.debug()
 
+Sentry.init({
+  dsn: 'https://examplePublicKey@o0.ingest.sentry.io/0',
+
+  // We recommend adjusting this value in production, or using tracesSampler
+  // for finer control
+  tracesSampleRate: 1.0,
```
marandaneto, line 33: I think you meant https://docs.sentry.io/platforms/node/configuration/options/#sample-rate instead?
`tracesSampleRate` is for transactions, `sampleRate` is for errors which is this integration for.

benjackwhite: I just copy pasted their default :D

posthog-node/src/extensions/sentry-integration.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Integrate Sentry with PostHog. This will add a direct link to the person in Sentry, and an $exception event in PostHog.
+ *
+ * ### Usage
+ *
+ *     Sentry.init({
+ *          dsn: 'https://example',
+ *          integrations: [
+ *              new posthog.SentryIntegration(posthog)
```
marandaneto, line 51:
```suggestion
 *              new posthog.PostHogSentryIntegration(posthog)
```

marandaneto (approved):
LGTM

benjackwhite:
> After this PR it'd be cool to have docs -> https://posthog.com/docs/libraries/node similar to https://posthog.com/docs/libraries/python#sentry so it's more discoverable.

Yep was just doing this :D

posthog-node/src/extensions/sentry-integration.ts
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      const sentry = getCurrentHub()
+
+      // Get the PostHog user ID from a specific tag, which users can set on their Sentry scope as they need.
+      const userId = event.tags[PostHogSentryIntegration.POSTHOG_ID_TAG]
+      if (userId === undefined) {
+        // If we can't find a user ID, don't bother linking the event. We won't be able to send anything meaningful to PostHog without it.
+        return event
+      }
```
pauldambra, line 96: I guess we could add a debug log here to help someone implementing it that didn't realise they have to set the tag... but they can also read this code so 🤷

pauldambra (approved):
looks good to me!

posthog-node/src/extensions/sentry-integration.ts
benjackwhite: True... But then it could also create a huge number of logs that you can't avoid (maybe you purposefully only set it for certain events...) Lets see if we actually get anyone using it 😓

Issue #143, opened 2023-12-19
React Native app tested with jest fails with: PostHog: No storage available. Please install expo-filesystem or react-native-async-storage OR implement a custom storage provider.


itsramiel:
## Bug description
When trying to test a react native app with PostHog installed, the following error is thrown:
`Error: PostHog: No storage available. Please install expo-filesystem or react-native-async-storage OR implement a custom storage provider. at buildOptimisiticAsyncStorage` even though async storage is installed

## How to reproduce

1. create a new react native project with `npx react-native@latest init AwesomeProject` as mentioned in the [docs](https://reactnative.dev/docs/environment-setup)
2. Install with `yarn add posthog-react-native @react-native-async-storage/async-storage react-native-device-info` and configure posthog
3. run `yarn test` -> Fails

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

benjackwhite:
To me it sounds like the error describes what needs to be done quite clearly 😅

What exactly is the bug?

itsramiel:
Installation was done with `yarn add posthog-react-native @react-native-async-storage/async-storage react-native-device-info`, so async storage is there. 

It works fine when running on a device or simulator, but not in jest.

What needs to be done?

itsramiel:
@benjackwhite  were you able to reproduce that?

marandaneto:
@itsramiel did you check that your libs are also available for the `devDependencies` target?
Try to run the very same command with the `--dev` argument, which then installs the libs as `devDependencies`.

itsramiel:
Hi @marandaneto. Yes that also doesnt work.

Here is a [repro](https://github.com/itsramiel/posthog-test-repro) with 4 branches installing as dep/devDep and mocking async storage vs not:

<img width="394" alt="image" src="https://github.com/PostHog/posthog-js-lite/assets/80689446/05e49fec-d2d3-4432-9a16-8e98a73b48a9">

Please let me know the issue. Or feel free to create a repo that successfully passes `yarn test` initialized with a new react native project for me to check out

marandaneto:
The problem is that those libraries don't support macOS/Linux/Windows, just the Android/iOS targets mostly.
So trying to load them via tests won't work.

You can do that (mocking) in many ways:

By setting a `customAsyncStorage` mock, or `persistence: 'memory'` during testing.

```
[...]

PR #146, opened 2023-12-21
If `async-storage` or `expo-file-system` is not installed, the SDK will fallback to `persistence: memory` and log a warning

marandaneto:
## Problem

If you install the PH SDK in a clean sample app and run `yarn test`, the test will fail because the PH SDK needs the storage libs but they aren't available for other targets than Android and iOS.
The only option is to mock the libs or set the persistence to `memory`, both are not great as a first experience when using our SDK so it's better to use a fallback instead of throwing an error and letting the user figure it out.

Fixes https://github.com/PostHog/posthog-js-lite/issues/143

## Changes

SDK fallback to `persistence: memory` if storage libs aren't installed.
Added a test but it failed because of https://github.com/facebook/react-native/issues/35701 so I removed it, I tested the code at runtime though.

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -37,7 +37,7 @@ class PostHogFetchNetworkError extends Error {
 
   constructor(public error: unknown) {
     // TRICKY: "cause" is a newer property but is just ignored otherwise. Cast to any to ignore the type issue.
-    // @ts-ignore
+    // @ts-expect-error
     super('Network error while fetching PostHog', error instanceof Error ? { cause: error } : {})
   }
 }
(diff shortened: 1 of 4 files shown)


benjackwhite (approved):
I really don't like this solution but I can't think of a better one so lets do it 😅

marandaneto:
> I really don't like this solution but I can't think of a better one so lets do it 😅

Same 😅

PR #149, opened 2023-12-26
feat(flags): Add local props and flags to all calls

neilkakkar:
## Problem

Copy of https://github.com/PostHog/posthog-python/pull/106/

This turned out to be a lot trickier than expected, because I noticed a few issues. With the `sendFeatureFlags` capture option, it seems like we were dropping events, since on shutdown the feature flag promises aren't tracked for populating flag information, which means the capture is not enqueued.

I can't yet think of a better way to solve this right now, so we just start tracking flag promises from within capture as well, annnd flush again on shutdown after pending promises have resolved, to make sure all new capture events queued are sent as well.

I'm also moving a few more things that should've been based on the onError emitter, and fixed tests that incorrectly assumed this was working fine.

(cc: @marandaneto @benjackwhite have a look here too please?)

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-node/src/posthog-node.ts
+++ b/posthog-node/src/posthog-node.ts
@@ -122,6 +122,26 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
         }
         _capture({ ...properties, $groups: groups, ...flagProperties })
       })
+    } else if (this.featureFlagsPoller?.featureFlags) {
+      const groupsWithStringValues: Record<string, string> = {}
+      for (const [key, value] of Object.entries(groups || {})) {
+        groupsWithStringValues[key] = String(value)
+      }
+      this.getAllFlags(distinctId, { groups: groupsWithStringValues, disableGeoip, onlyEvaluateLocally: true }).then(flags => {
+        const featureVariantProperties: Record<string, string | boolean> = {}
+        if (flags) {
+          for (const [feature, variant] of Object.entries(flags)) {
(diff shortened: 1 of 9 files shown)


neilkakkar:
Missing a test for adding default properties to all flag functions

posthog-node/src/posthog-node.ts
```diff
@@ -106,22 +107,68 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
...
-          for (const [feature, variant] of Object.entries(flags)) {
-            if (variant !== false) {
-              featureVariantProperties[`$feature/${feature}`] = variant
+      const promiseUUID = generateUUID()
+      // :TRICKY: If we flush, or need to shut down, to not lose events we want this promise to resolve before we flush
+      const promise = super.getFeatureFlagsStateless(distinctId, groups, undefined, undefined, disableGeoip)
+      this.pendingPromises[promiseUUID] = promise
+
+      promise
+        .then((flags) => {
```
neilkakkar, line 116: oh, add cleanup to delete this promise from here too, once resolved

neilkakkar:
oh much nicer, thanks 😓

jurajmajerik (approved):

PR #150, opened 2024-01-05
fix: local props refactor

benjackwhite:
## Problem

Made some changes but pulled them to a PR as I'm not so sure about them

@neilkakkar:
* Changes some of the expectations so that the payload for feature flag properties is always the same, regardless of source - worth checking if this is correct
* Fixes a bad issue with the `waitForPromises` that probably caused a lot of confusion elsewhere so hopefully thats an improvement
* This makes every `capture` call a promise but simplifies the logic and makes it more reliable which feels better


neilkakkar:
works great, just tested it, thanks!

PR #156, opened 2024-01-18
feat: Add eventId overriding options

benjackwhite:
## Problem

Some people want to override the eventId to do things like deduplication

## Changes

* Adds a default uuid
* Accepts the override option

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [x] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added the ability to override the event ID via the extra options argument in capture calls


marandaneto (approved):
LGTM

PR #160, opened 2024-01-23
feat: Move to uuidv7

benjackwhite:
## Problem

We moved to uuidv7 on posthog-js - maybe we should do the same for js-lite?

## Changes

* Moves the utils around and swaps to uuidv7 with the same approach as posthog-js

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

* Swaps the internal uuid implementation to uuidv7

- [x] All of them
- [x] posthog-web
- [x] posthog-node
- [x] posthog-react-native

### Changelog notes

- Added support for X


marandaneto (approved):
Just changelog/version missing, otherwise LGTM

.github/workflows/bundle-size.yml
pauldambra, line 1: nice 🙌

pauldambra (approved):
we were pretty careful when moving posthog-js across, so assuming you've done a sanity check of the uuids being generated here then I think 🚢 like a 🚀

marandaneto:
@benjackwhite should we merge this? (changelogs need fixing)

benjackwhite:
> @benjackwhite should we merge this? (changelogs need fixing)

Too many other things going on. Feel free to take over and merge if you are happy with it

PR #162, opened 2024-01-23
chore: React Native for macOS support

marandaneto:
### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/src/optional/OptionalExpoFileSystem.ts
+++ b/posthog-react-native/src/optional/OptionalExpoFileSystem.ts
@@ -4,12 +4,12 @@ import { Platform } from 'react-native'
 export let OptionalExpoFileSystem: typeof ExpoFileSystem | undefined = undefined
 
 try {
-  // do not try to load expo-file-system on web, otherwise it will throw an error
+  // do not try to load expo-file-system on web and macos, otherwise it will throw an error
   // Error: The method or property expo-file-system.writeAsStringAsync is not available on web
   // See https://github.com/PostHog/posthog-js-lite/issues/140
-  // Once expo-file-system is supported on web, we can remove this try/catch block
+  // Once expo-file-system is supported on web/macos, we can remove this try/catch block
   // For now, use the react-native-async-storage/async-storage package instead
-  if (Platform.OS !== 'web') {
+  if (Platform.OS !== 'web' && Platform.OS !== 'macos') {
     OptionalExpoFileSystem = require('expo-file-system') // No Web support
   }
 } catch (e) {}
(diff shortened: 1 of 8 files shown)


examples/example_rn_macos/App.tsx
```diff
@@ -0,0 +1,137 @@
...
+  LearnMoreLinks,
+  ReloadInstructions,
+} from 'react-native/Libraries/NewAppScreen';
+
+import PostHog from 'posthog-react-native';
+
+export let posthog: PostHog | undefined;
+
+export const posthogAsync: Promise<PostHog> = PostHog.initAsync(
+  'phc_pQ70jJhZKHRvDIL5ruOErnPy6xiAiWCqlL4ayELj4X8',
```
daibhin, line 33: Should we exclude this or change it to an env variable? Would not like the source to be polluted with data if people run the example

posthog-react-native/src/optional/OptionalExpoFileSystem.ts
```diff
@@ -4,12 +4,12 @@ import { Platform } from 'react-native'
...
-  // do not try to load expo-file-system on web, otherwise it will throw an error
+  // do not try to load expo-file-system on web and macos, otherwise it will throw an error
   // Error: The method or property expo-file-system.writeAsStringAsync is not available on web
   // See https://github.com/PostHog/posthog-js-lite/issues/140
-  // Once expo-file-system is supported on web, we can remove this try/catch block
+  // Once expo-file-system is supported on web/macos, we can remove this try/catch block
   // For now, use the react-native-async-storage/async-storage package instead
-  if (Platform.OS !== 'web') {
+  if (Platform.OS !== 'web' && Platform.OS !== 'macos') {
     OptionalExpoFileSystem = require('expo-file-system') // No Web support
```
daibhin, line 13:
```suggestion
    OptionalExpoFileSystem = require('expo-file-system') // No Web or MacOS support
```

examples/example_rn_macos/App.tsx
marandaneto: it's just an example project with a testing org, it's nice to see events coming through, otherwise every time I run the examples, I have to change the apiKey before, rolling back the changes, its working quite well on the other SDKs as well but its only me testing it anyway.

posthog-react-native/src/optional/OptionalExpoFileSystem.ts
marandaneto: will actually remove the comment since I mentioned above:

> do not try to load expo-file-system on web and macos

Issue #164, opened 2024-01-25
RN: expo-file-system could not be found within the project or in these directories (posthog-react-native >= 2.10.2)


leeseonghyun21:
## Bug description
In a bare React Native app environment, the following error occurs during Metro bundling when only using [@react-native-async-storage/async-storage](https://github.com/react-native-async-storage/async-storage). This issue occurs in a non-Web environment and when expo-file-system is not installed.

```sh
BUNDLE  ./index.js

error: Error: Unable to resolve module expo-file-system from /Users/dev/sample-project-name/node_modules/posthog-react-native/lib/posthog-react-native/src/optional/OptionalExpoFileSystem.js: expo-file-system could not be found within the project or in these directories:
  node_modules
  ../node_modules
  ../../node_modules
> 1 | "use strict";Object.defineProperty(exports,"__esModule",{value:true});exports.OptionalExpoFileSystem=void 0;var _reactNative=require("react-native");var OptionalExpoFileSystem=undefined;exports.OptionalExpoFileSystem=OptionalExpoFileSystem;try{if(_reactNative.Platform.OS!=='web'){exports.OptionalExpoFileSystem=OptionalExpoFileSystem=require('expo-file-system');}}catch(e){}
    |                                                                                                                                                                                                                                                                                                                                                         ^
    at ModuleResolver.resolveDependency (/Users/dev/sample-project-name/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:153:15)
    at DependencyGraph.resolveDependency (/Users/dev/sample-project-name/node_modules/metro/src/node-haste/DependencyGraph.js:271:43)
    at /Users/dev/sample-project-name/node_modules/metro/src/lib/transformHelpers.js:176:21
    at resolveDependencies (/Users/dev/sample-project-name/node_modules/metro/src/DeltaBundler/buildSubgraph.js:56:25)
    at visit (/Users/dev/sample-project-name/node_modules/metro/src/DeltaBundler/buildSubgraph.js:107:30)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Promise.all (index 8)
    at async visit (/Users/dev/sample-project-name/node_modules/metro/src/DeltaBundler/buildSubgraph.js:116:5)
    at async Promise.all (index 12)
    at async visit (/Users/dev/sample-project-name/node_modules/metro/src/DeltaBundler/buildSubgraph.js:116:5)
```
The operating environment is as follows.
```sh
System:
[...]

PR #165, opened 2024-01-25
fix: changed platform detection method to select

leeseonghyun21:
## Problem
This PR fixes https://github.com/PostHog/posthog-js-lite/issues/164

## Changes
changed the logic for checking the Platform from the original `Platform.OS` + `if` statement to `Platform.select`.

## Release info Sub-libraries affected
nth
### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native


marandaneto:
@leeseonghyun21 Thanks for the PR, I will close this one in favor of https://github.com/PostHog/posthog-js-lite/pull/166 since I had to change more files, I was not aware of the `Platform.select` approach, very handy and surprised that it behaves differently than the conditional require.

PR #170, opened 2024-01-31
feat: support sync provider in rn

filipef101:
### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/storage-memory.ts
+++ b/posthog-core/src/storage-memory.ts
@@ -1,13 +1,13 @@
-import { PostHogPersistedProperty } from './types'
+import { PostHogPersistedProperty } from "./types";
 
 export class PostHogMemoryStorage {
-  private _memoryStorage: { [key: string]: any | undefined } = {}
+  private _memoryStorage: { [key: string]: any | undefined } = {};
 
-  getProperty(key: PostHogPersistedProperty): any | undefined {
-    return this._memoryStorage[key]
+  getItem(key: PostHogPersistedProperty): any | undefined {
+    return this._memoryStorage[key];
   }
 
-  setProperty(key: PostHogPersistedProperty, value: any | null): void {
-    this._memoryStorage[key] = value !== null ? value : undefined
+  setItem(key: PostHogPersistedProperty, value: any | null): void {
+    this._memoryStorage[key] = value !== null ? value : undefined;
   }
 }
--- a/posthog-node/src/posthog-node.ts
+++ b/posthog-node/src/posthog-node.ts
@@ -1,4 +1,4 @@
-import { version } from '../package.json'
+import { version } from "../package.json";
 
 import {
   JsonType,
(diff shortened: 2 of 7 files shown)


filipef101:
quick on-the-keen proof of concept mostly atm

posthog-react-native/src/storage.ts
```diff
@@ -62,20 +74,11 @@ export class SemiAsyncStorage {
...
+    if (value === null) {
       delete this._memoryCache[key]
+    } else {
+      this._memoryCache[key] = value
     }
     this.persist()
   }
-  getAllKeys(): readonly string[] {
-    return Object.keys(this._memoryCache)
-  }
```
filipef101, line 80: not used

posthog-core/src/storage-memory.ts
```diff
@@ -3,11 +3,11 @@ import { PostHogPersistedProperty } from './types'
...
   private _memoryStorage: { [key: string]: any | undefined } = {}
 
-  getProperty(key: PostHogPersistedProperty): any | undefined {
+  getItem(key: PostHogPersistedProperty): any | undefined {
     return this._memoryStorage[key]
   }
 
-  setProperty(key: PostHogPersistedProperty, value: any | null): void {
+  setItem(key: PostHogPersistedProperty, value: any | null): void {
     this._memoryStorage[key] = value !== null ? value : undefined
```
filipef101, line 11: renamed for consistency

posthog-node/src/posthog-node.ts
```diff
@@ -178,7 +182,9 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
   }
 
   alias(data: { distinctId: string; alias: string; disableGeoip?: boolean }): void {
-    super.aliasStateless(data.alias, data.distinctId, undefined, { disableGeoip: data.disableGeoip })
+    super.aliasStateless(data.alias, data.distinctId, undefined, {
+      disableGeoip: data.disableGeoip,
+    })
```
filipef101, line 187: prettier 🤷

posthog-react-native/src/posthog-rn.ts
```diff
@@ -10,30 +10,31 @@ import {
...
+  /** If you use a customProvider that is synchronous, ie memory, JSI, cookies/localStorage on rn-web.
+   * You may `customSync` so posthog gets initialized with `new PostHog(..`
+   * */
   persistence?: 'memory' | 'file'
   /** Allows you to provide your own implementation of the common information about your App or a function to modify the default App properties generated  */
   customAppProperties?:
     | PostHogCustomAppProperties
     | ((properties: PostHogCustomAppProperties) => PostHogCustomAppProperties)
-  customAsyncStorage?: PostHogCustomAsyncStorage
+  customAsyncStorage?: PostHogStorage
```
filipef101, line 28: customAsyncStorage only needs to implement a subset of the posthog one, not sure if anyone would even use this

posthog-react-native/src/posthog-rn.ts
```diff
@@ -10,30 +10,31 @@ import {
...
 }
 
 const clientMap = new Map<string, Promise<PostHog>>()
 
 export class PostHog extends PostHogCore {
-  private _persistence: PostHogOptions['persistence']
-  private _memoryStorage = new PostHogMemoryStorage()
-  private _semiAsyncStorage?: SemiAsyncStorage
+  private _persistence: 'sync' | 'async' | 'memory'
+  private _storage: SemiAsyncStorage | PostHogMemoryStorage
```
filipef101, line 36: unify storage property

posthog-react-native/src/posthog-rn.ts
```diff
@@ -87,31 +102,32 @@ export class PostHog extends PostHogCore {
...
+        storage || new PostHogSemiAsyncStorage(options?.customAsyncStorage || buildOptimisiticAsyncStorage())
     }
 
     // Ensure the async storage has been preloaded (this call is cached)
-
-    const setupAsync = async (): Promise<void> => {
-      if (!this._semiAsyncStorage?.isPreloaded) {
-        await this._semiAsyncStorage?.preloadAsync()
-      }
+    const setup = () => {
```
filipef101, line 122: moved the await for preLoadAsync to the initAsync, so it also gets called on the constructor for non async storage (memory or customSync).
initAsync should be called for async storage, ie default. But should not, for sync, ie memory, custom storage passed to new PostHog(...., fancyStorage)

posthog-react-native/src/posthog-rn.ts
```diff
@@ -125,26 +141,18 @@ export class PostHog extends PostHogCore {
...
-      ? this._semiAsyncStorage.getItem(key) ?? undefined
-      : this._memoryStorage.getProperty(key)
+    return this._storage.getItem(key) ?? undefined
   }
   setPersistedProperty<T>(key: PostHogPersistedProperty, value: T | null): void {
-    return this._semiAsyncStorage
-      ? value !== null
-        ? this._semiAsyncStorage.setItem(key, value)
-        : this._semiAsyncStorage.removeItem(key)
-      : this._memoryStorage.setProperty(key, value)
```
filipef101, line 147: moved this hacky-ish piece of code to storage.ts

posthog-react-native/src/storage.ts
```diff
@@ -1,22 +1,34 @@
...
+import { PostHogStorage } from './types'
 
 const POSTHOG_STORAGE_KEY = '.posthog-rn.json'
 const POSTHOG_STORAGE_VERSION = 'v1'
 
 type PostHogStorageContents = { [key: string]: any }
 
-// NOTE: The core prefers a synchronous storage so we mimic this by pre-loading all keys
-export class SemiAsyncStorage {
+export abstract class SemiAsyncStorage {
```
filipef101, line 8: This is what a custom "semi async storage" should implement, and what the now called PostHogSemiAsyncStorage implements

filipef101:
Also no idea if TS will scream, didn't work with typescript enabled

posthog-react-native/src/posthog-rn.ts
```diff
@@ -87,31 +102,32 @@ export class PostHog extends PostHogCore {
...
-      if (!this._semiAsyncStorage?.getItem(PostHogPersistedProperty.AnonymousId)) {
+      if (!this._storage?.getItem(PostHogPersistedProperty.AnonymousId)) {
         const legacyValues = await getLegacyValues()
         if (legacyValues?.distinctId) {
-          this._semiAsyncStorage?.setItem(PostHogPersistedProperty.DistinctId, legacyValues.distinctId)
-          this._semiAsyncStorage?.setItem(PostHogPersistedProperty.AnonymousId, legacyValues.anonymousId)
+          this._storage?.setItem(PostHogPersistedProperty.DistinctId, legacyValues.distinctId)
+          this._storage?.setItem(PostHogPersistedProperty.AnonymousId, legacyValues.anonymousId)
         }
       }
```
filipef101, line 132: This actually is the only blocker, no idea about it, maybe we can move this to the initAsync? or could we do a .then instead of await?

marandaneto:
Thanks @filipef101 
I will review it early next week.

posthog-react-native/src/posthog-rn.ts
marandaneto: The problem with `then` is that it might be executed a bit after and `getItem` may be called before the `then` gets the chance to persist the data and read outdated data.

marandaneto:
@filipef101 Sorry for the late review, and thanks for the PR.
I think it'd make more sense to refactor this with proper sync/async support in mind and cut a v3 of the RN SDK (less breaking changes as possible but not avoiding it to make it better).

Most likely the SDK should offer a full sync and/or async approach by exposing both options, the semi-async approach makes it a bit more difficult to keep the code clean and concise.

It'd be better to expose sync and async methods for everything IMO, similar to `flush` and `flushAsync`, so we don't have to abstract so much to make things work, but the client rather inits and calls the methods based on their needs, either that be sync or async storages.

Your approach is almost there, I will apply some changes within the next days or weeks depending on how much time I can invest in this.

filipef101:
Yup, this is more a proposal. Good work!

filipef101:
Feel free to close if you want

PR #172, opened 2024-02-07
chore: add device model and replace device name

marandaneto:
## Problem

Noticed by checking the `most common devices` dashboard template that RN used the wrong property for device model and name

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- Added support for X


benjackwhite (commented):
Code-wise no issues, but is this technically a breaking change? Not so much for the SDK but for the analytics data on the other end 🤔

posthog-react-native/src/native-deps.tsx
```diff
@@ -40,12 +41,13 @@ export const getAppProperties = (): PostHogCustomAppProperties => {
   }
 
   if (OptionalReactNativeDeviceInfo) {
-    properties.$app_build = returnPropertyIfNotUnknown(OptionalReactNativeDeviceInfo.getBuildIdSync())
+    properties.$app_build = returnPropertyIfNotUnknown(OptionalReactNativeDeviceInfo.getBuildNumber())
```
marandaneto, line 44: This matches the right app build when using expo-device and the Android/iOS SDKs
getBuildId != BuildNumber

benjackwhite (approved):

marandaneto:
the RN build command is not packing the `uuidv7` dep so I have to fix that first.

posthog-react-native/src/native-deps.tsx
marandaneto: it was returning `UP1A.231105.003` instead of `1` which is the app build

marandaneto:
close in favor of https://github.com/PostHog/posthog-js-lite/pull/185

PR #173, opened 2024-02-08
Fix: [React Native] - Vendor uuidv7 instead of using peer dependency to avoid the missing crypto issue

marandaneto:
## Problem

The RN publish cmd isn't copying the peer dependencies correctly.
After fixing this, there was an issue with the `crypto` package which is `node` only.
Loading packages dynamically on RN is better with `Platform.select` otherwise it causes issues on macOS/Web, etc.
`Platform.select` is only available on React Native, so we cannot use it on the core package.
The best solution was to vendor the `uuidv7` which is apache2 anyway and let `getDefaultRandom` use only the `Math.random` available on all platforms.

Only affects RN because `crypto` was available on the other env.

## Changes

Vendor `uuidv7`

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/vendor/uuidv7.ts
+++ b/posthog-core/src/vendor/uuidv7.ts
@@ -1,0 +1,478 @@
+// vendor from: https://github.com/LiosK/uuidv7/blob/f30b7a7faff73afbce0b27a46c638310f96912ba/src/index.ts
+
+/**
+ * uuidv7: An experimental implementation of the proposed UUID Version 7
+ *
+ * @license Apache-2.0
+ * @copyright 2021-2023 LiosK
+ * @packageDocumentation
+ */
+
+const DIGITS = "0123456789abcdef";
+
+/** Represents a UUID as a 16-byte byte array. */
+export class UUID {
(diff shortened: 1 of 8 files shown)


posthog-core/src/vendor/uuidv7.ts
```diff
@@ -0,0 +1,478 @@
...
+//     // fall back on Math.random() unless the flag is set to true
+//     if (typeof UUIDV7_DENY_WEAK_RNG !== "undefined" && UUIDV7_DENY_WEAK_RNG) {
+//       throw new Error("no cryptographically strong RNG available");
+//     }
+//     return {
+//       nextUint32: (): number =>
+//         Math.trunc(Math.random() * 0x1_0000) * 0x1_0000 +
+//         Math.trunc(Math.random() * 0x1_0000),
+//     };
+//   }
```
marandaneto, line 429: important part is here

benjackwhite (approved):

Issue #176, opened 2024-02-12
Improve offline support


marandaneto:
### Description

Context https://posthog.com/questions/offline-usage-1

It partially does. It persists the queue. The only thing is, that it doesn't:

> pause flushing if the network is unreachable

this is nice to have since network connectivity callbacks only tell there's a cellphone signal or wifi connected but not necessarily the internet.

> add messages back to the queue if the flush fails

this is a must, if an app is offline first, and it's always offline, the queue will try to be flushed and will remove the items from the queue, but it will fail since there's no connection at all, that slice of the queue will be lost if not restored.

johnslemmer:
Hey @marandaneto I noticed this got closed with #177. Wondering if you can provide any more context on how offline support is improved.  Is the queue now cached on the file system?  Is there any limitation to the number events that this queue can store or length of time that an app can be offline for?

Thanks for working on this!!

marandaneto:
All events will be cached if the device is offline, there's a timer flushing the events to the network, and the timer is configurable via `flushInterval`.
If the device is offline, the events will remain cached, once the timer is triggered and there's network, the events will be sent and deleted from the local storage.
Previously the events were deleted from the local storage even if flusing failed, so events were lost anyways.

rafaellehmkuhl:
It is important to mention that for the queue to be persistent between sessions (e.g.: the app is used offline several times, being opened and closed, and at some point it is used online), the user must set the `persistence` key to `localStorage` in the initial configuration:

```javascript
PostHog.posthog = posthog.init('your_key', {
  api_host: 'https://us.i.posthog.com',
  persistence: 'localStorage',
  persistence_name: 'posthog-persistent-queue-or-the-name-you-prefer'
})
```

PR #177, opened 2024-02-12
feat: Improved queue flushing

marandaneto:
## Problem

Fixes https://github.com/PostHog/posthog-js-lite/issues/176

## Changes

* Flushing will only remove from the queue if it has reliably been dealt with
* Only one flush can happen at a time but with an increased `maxBatchSize`
* Shutting down works as designed, iteratively calling flush until the queue is exhausted or the timeout is reached.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [x] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X


benjackwhite (changes requested):
What happens if you flush twice in a row? Feels like we are missing something to block multiple flushes (or at least handle them correctly). 

Similarly we might want to check that the shutdown logic would still work if we block multiple flushes at once

marandaneto:
> What happens if you flush twice in a row? Feels like we are missing something to block multiple flushes (or at least handle them correctly).
> 
> Similarly we might want to check that the shutdown logic would still work if we block multiple flushes at once

True, I assumed the method was `sync` but there's a `then` call at the end, and that would lead to race conditions and messing up the queue, good catch.
Only 1 flush per time is allowed to avoid such races using the state flag `isFlushing`.
The `shutdown` logic should be fine because it [awaits all pending promises](https://github.com/PostHog/posthog-js-lite/blob/98b3be5cdb0c933469b6602172ace950b2ed993b/posthog-core/src/index.ts#L600) anyway and call [flush](https://github.com/PostHog/posthog-js-lite/blob/98b3be5cdb0c933469b6602172ace950b2ed993b/posthog-core/src/index.ts#L609) again.
WDYT? can write tests if you think this is enough now.

posthog-core/src/index.ts
```diff
@@ -456,6 +457,12 @@ export abstract class PostHogCoreStateless {
   }
 
   flush(callback?: (err?: any, data?: any) => void): void {
+    if (this.isFlushing) {
+      console.log('[PostHog] The Queue is already flushing.')
+      return
```
benjackwhite, line 462: Hmmmm the callback is important here otherwise the async calls never fire... 
If we add the callback it becomes tricky as it means the flushAsync will resolve immediately which is not really true 🤔 

We could do a promise-hold approach perhaps instead which would give us a way of tracking the inflight flush?

benjackwhite (changes requested):
Happy to jump in and help here as this is a bit more JS-y than mobile-y

posthog-core/src/index.ts
benjackwhite: This is made more important by the fact it affects Node as well

marandaneto: yeah should have called `callback?.()` here, and pushed a change just so that the `flushAsync` works as it should.

> flushAsync will resolve immediately which is not really true 🤔

yep but it's fine I think, I'd consider that a best effort, the `_flushTimer` will get triggered again in the next cycle.

The only issue I see is the `shutdownAsync`, and `flushAsync` will return right away but even with that, it's not guaranteed that the whole Queue will be flushed even after calling `flushAsync` twice, so if the flush was on going during the `shutdownAsync` call, I'd consider that ok and send the rest next time the SDK is up and running again.
If the Queue has to be fully flushed before shutdown sync is completed, even if the current implementation does not work, we'd need to iterate the Queue until it's empty.

marandaneto:
> What happens if you flush twice in a row? Feels like we are missing something to block multiple flushes (or at least handle them correctly).
> 
> Similarly we might want to check that the shutdown logic would still work if we block multiple flushes at once

The `flush` method only calls `removeItemsFromQueue` and it's a sync method and JS is single-threaded.
In React Native land we could use `useState` to be sure and avoid race conditions, but this method isn't async so I guess this is fine?


> Happy to jump in and help here as this is a bit more JS-y than mobile-y

yep, made a new comment [here](https://github.com/PostHog/posthog-js-lite/pull/177#discussion_r1486105310), but happy to hear your thoughts since this has node/web impact.

promise-hold approach would work but I wonder if it's really necessary, I'd say it's only fully necessary if we want close to real-time events, where any call to flush or flushAsync cannot be missed and cannot wait the next timer cycle.
The only downside is the `shutdownAsync`, but as mentioned in my comment, even with the current implementation, the Queue isn't fully flushed.

posthog-core/src/index.ts
marandaneto: discussed offline, ideally, the shutdown should not bail if it's already flushing but rather awaits the pending flushes and flushes everything else as well until the very end (iterate the queue), this is important for node, specifically serverless/cloud functions stuff that boots and dies right away, all the pending events has to be flushed otherwise they are lost.

posthog-core/src/index.ts
```diff
@@ -52,7 +52,9 @@ export abstract class PostHogCoreStateless {
   private apiKey: string
   host: string
   private flushAt: number
+  private flushMaxItems: number
```
marandaneto, line 55: I call it `maxBatchSize` on Mobile, not sure if relevant to unify though.

posthog-core/src/types.ts
```diff
@@ -43,6 +43,7 @@ export enum PostHogPersistedProperty {
   FeatureFlagPayloads = 'feature_flag_payloads',
   OverrideFeatureFlags = 'override_feature_flags',
   Queue = 'queue',
+  FlushQueue = 'flush_queue',
```
marandaneto, line 46: We don't use that, right? we can remove it.

posthog-core/src/index.ts
```diff
@@ -491,97 +498,123 @@ export abstract class PostHogCoreStateless {
...
-        return err ? reject(err) : resolve(data)
+  /**
+   * Helper for flushing the queue in the background
+   * Avoids unnecessary promise errors
+   */
+  private flushBackground(): void {
+    void this.flush().catch(() => {})
+  }
+
+  async flush(): Promise<any[]> {
```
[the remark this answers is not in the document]
benjackwhite: good call

benjackwhite (approved):

Issue #114, opened 2023-10-03
Don't complain about "initAsync called twice" when hot reloading


statico:
## Is your feature request related to a problem?

Similar to #113 — every time my React Native app is loaded, reloaded, or the base layout changes (we're using expo-router), PostHog throws up a giant warning:

> PostHog.initAsync called twice with the same apiKey. The first instance will be used.

It would be nice to disable this message when `__DEV__` is true.

Recoil, a state management library for React, had a similar behavior with singletons and hot reloading and solved this by allowing developers to add a `RECOIL_DUPLICATE_ATOM_KEY_CHECKING_ENABLED=false` setting: https://github.com/facebookexperimental/Recoil/issues/733

## Describe the solution you'd like

Some kind of setting, maybe in PostHogProvider, like `dangerouslyIgnoreReinitialization` or `iKnowWhatImDoing={__DEV__}` or something, that removes this warning.

## Describe alternatives you've considered

- Patching the package with `patch-package`

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

## Additional context

<3

hikchoi:
seconding this. pretty harmless, but it makes it hard to read other logs that I actually want to read :(

@statico are you still patching the package on your end or have you found any other workarounds?

statico:
@hikchoi Yep, still patching this and a few other messages. Previously with `patch-package`, now with pnpm's built-in patcher.

marandaneto:
Maybe the solution here is to patch the Router and not call `PostHog.initAsync` again when a hot reload just happened, you can do that via env. var. (similar to [113](https://github.com/PostHog/posthog-js-lite/issues/113)) or a global variable that survives hot reload.

We already do that [here](https://github.com/PostHog/posthog-js-lite/blob/a604f6a9b62c8a8b351b23797bda6566ec0366d0/posthog-react-native/src/PostHogProvider.tsx#L68-L69) but if the whole Provider gets recreated, this field is also gone, the solution would be to depend on an external global variable.

edumana:
I was getting this error and solved it by putting PH provider at the very top of the `RootLayout`, and **moving everything to a new <App/> child component**. I had some auth states changing which were causing PH provider to reload. So my app now looks like this:

`const RootLayout = () => {
  return (
      <PostHogProvider
        autocapture={{
          captureScreens: false,
        }}
[...]

PR #179, opened 2024-02-13
chore: remove PostHog.initAsync warning

zakharchenkoAndrii:
## Problem

The recommended way to set up PostHog for React Native is to use the PostHogProvider. Indeed that is true
However while using it like that we are facing console warning which annoys and distracts from really important things.

Folks [here](https://github.com/PostHog/posthog-js-lite/issues/114) already reported about this problem

## Changes

Just removed console.warn. It more harms than helps, you already did a great job to prevent re-initialization, so I don't think that warning helps

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- Removed "PostHog.initAsync called twice with the same apiKey" console warning

### DIFF
--- a/posthog-react-native/src/posthog-rn.ts
+++ b/posthog-react-native/src/posthog-rn.ts
@@ -64,8 +64,6 @@ export class PostHog extends PostHogCore {
       }
 
       clientMap.set(apiKey, posthog)
-    } else {
-      console.warn('PostHog.initAsync called twice with the same apiKey. The first instance will be used.')
     }
 
     const resolved = await posthog


benjackwhite (approved):
Seems fair enough. If we had a proper debug logger, I would recommend that we log it there so that it shows up as warning with debugging enabled but for now this works

marandaneto:
Yeah the warning isn't great but it's possible to avoid the recreation every single time as well depending on where/how you setup the provider, see comment https://github.com/PostHog/posthog-js-lite/issues/114#issuecomment-1892624084
the logs should be behind a flag called `debug` so it'd only be there if it's enabled, since now everything outputs to the console, it's more harmful than useful, agreed.

marandaneto (approved):

marandaneto:
changed the level to patch instead of major, its just a warning log.

marandaneto:
Thanks @zakharchenkoAndrii

benjackwhite:
Just a reminder that the bumping needs to be done manually in this repo

Issue #180, opened 2024-02-13
Bug: posthog.on('error') always catches http fetch error


L03TJ3:
## Bug description
We added a logger to catch whenever posthog failed to load at all. It seems posthog.on('error') should be used for this.
There is an initial HTTP fetch error always logged and catched, even if the calls to host are succesfull.

## How to reproduce
1. Not sure. pretty straigh-forward implementation where we only use getFeatureFlag(s)
2. when relying on the default where sendFeatureFlagEvent is enabled, the .on('error') always triggers after default retry's and timeouts

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

## Additional context
1. we use a reverse proxy
2. it seems to be triggered once sendFeatureFlagEvent is enabled. disabling this and I don't catch a fetch error (which it is by default)

sirpy:
Also i've noticed that the fetch error is:
1. displayed with error code 0 (ie not an http error code)
2. I've noticed it happens when moving to another tab and going back
3. it doesnt seem to generate any network calls

marandaneto:
@sirpy can you enable `debug` and check the console logs? what is the error about? is there any more info either before or after the error?

Do first thing after the SDK is init:
`posthog.debug(false)`

I cannot reproduce it locally so I'm not sure what it can be.

sirpy:
@marandaneto 
We are using your service I believe US cloud. we use vercelrp.gooddollar.org as a reverse proxy to your api host.
Also notice that we are using the posthog react native in browser env.

marandaneto:
<img width="577" alt="Screenshot 2024-02-15 at 11 35 23" src="https://github.com/PostHog/posthog-js-lite/assets/5731772/e6c01f38-1e24-4d17-9ece-8e74babc40ff">

I can reproduce it using RN Web, will check it out, thanks.

marandaneto:
It's not really because of `sendFeatureFlagEvent`, if you call `posthog.capture('test');` you can reproduce that already, it just happens that `sendFeatureFlagEvent=true` captures an event when feature flags are used.

marandaneto:
If you init the SDK with `captureMode: 'json'`, this does not happen either, but its `form` by default.

marandaneto:
[...]

PR #181, opened 2024-02-15
fix: using `captureMode=form` won't throw an error and retry unnecessarily

marandaneto:
## Problem

Fixes https://github.com/PostHog/posthog-js-lite/issues/180

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [X] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X


posthog-core/src/index.ts
```diff
@@ -550,7 +550,11 @@ export abstract class PostHogCoreStateless {
           // fetch will only throw on network errors or on timeouts
           throw new PostHogFetchNetworkError(e)
         }
-        if (res.status < 200 || res.status >= 400) {
+        // If we're in no-cors mode, we can't access the response status
+        // We only throw on HTTP errors if we're not in no-cors mode
+        // https://developer.mozilla.org/en-US/docs/Web/API/Request/mode#no-cors
+        const isNoCors = options.mode === 'no-cors'
+        if (!isNoCors && (res.status < 200 || res.status >= 400)) {
```
marandaneto, line 557: important bits are here, we were checking the status although it was always 0, we were retrying 3 times every time, events were not duplicated because it was the very same id.

benjackwhite (approved):
Looks good i think!

Issue #182, opened 2024-02-16
React native app crashes during startup


johnslemmer:
## Bug description

First, everything works fine in development.  I build my own expo dev client app for development.  However, after I just made an official build of our app and install via Apple's TestFlight the app immediately crashes and I get this error in my Sentry logs:

```
TypeError
undefined is not a function

/Users/expo/workingdir/build/node_modules/posthog-react-native/lib/posthog-react-native/src/hooks/useLifecycleTracker.js in anonymous at line 1:520

{snip} kedRef.current){openTrackedRef.current=true;posthog.capture('Application Opened');}var subscription=_reactNative.AppState.addEventListener(' {snip}
```

I did dig through the PostHog code and saw that this will never run while the posthog client is undefined. 
 See https://github.com/PostHog/posthog-js-lite/blob/2911a8fbc17db974ec5ed2913a47f2b52bc1586b/posthog-react-native/src/hooks/useLifecycleTracker.ts#L12 .  And plus I wouldn've gotten a different error (eg. `Uncaught TypeError: Cannot read properties of undefined`). So my assumption is that `capture` is undefined somehow causing this error to happen, but that doesn't really make sense as posthog is a class instance and `capture` is defined on that class.  Any ideas?

Picture for reference:

<img width="1120" alt="image" src="https://github.com/PostHog/posthog-js-lite/assets/18610918/2a967ca2-b065-4033-91c2-190d05b9890f">

## How to reproduce

1. Setup up posthog "without the posthog provider" https://posthog.com/docs/libraries/react-native#without-the-posthogprovider but still using PosthogProvider.  Example setup:

```js
import { PostHog } from "posthog-react-native";

export const analytics = PostHog.initAsync(process.env.POSTHOG_API_KEY, {
  host: "https://app.posthog.com",
});

let analyticsSync: PostHog | undefined;
analytics.then((res) => (analyticsSync = res)).catch(console.error);

// wrap your app with this
export function AnalyticsProvider({
  children,
}: {
  children: React.ReactNode;
}) {
  return (
    <PostHogProvider
      client={analytics}
      autocapture={{
        captureTouches: true,
        captureLifecycleEvents: true,
        captureScreens: true,
		...
```

2. Build a standalone version of the iOS app (not a dev client).
3. Install on device and open.

I understand this isn't a full recreation.  I could try and spend some time recreating this with a simpler repo but it would have to wait a few days.  Let me know.

## Related sub-libraries

- [ ] All of them
[...]

PR #184, opened 2024-02-20
fix: undefined posthog in hooks

marandaneto:
## Problem

Probably solves https://github.com/PostHog/posthog-js-lite/issues/182

I suspect that when there are more promises to be resolved, posthog tries to use its instance but it's not ready yet.

## Changes

The changes allow passing the custom `posthog` to the hooks instead of always using `usePostHog`, which is likely already resolved since it was either passed or created by the provider.
If it's not ready, at least it'll bail out instead of crashing.

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X


daibhin (approved):
Change looks good to me. Not really clear what I would change in my code as an integrator to fix the crash (if anything). Might be worth documenting that or making an example if you think it's necessary

marandaneto:
> Change looks good to me. Not really clear what I would change in my code as an integrator to fix the crash (if anything). Might be worth documenting that or making an example if you think it's necessary

there's no change to be made since the `client` is optional and the provider creates them automatically, the provider already has an optional `client` which is already passed.

Issue #188, opened 2024-02-23
Events should be buffered until the async init is resolved and the SDK is able to capture events


marandaneto:
### Description

https://github.com/PostHog/posthog-js-lite/issues/182#issuecomment-1957534606
If RN is busy resolving lots of promises, some integrations such as `useLifecycleTracker` and others might bail out because `posthog` is yet not resolved.
The goal is that those events are queued in a map with <timestamp, event> and later sent to the network in order (FIFO).

This also relates to https://github.com/PostHog/posthog-js-lite/issues/96 because it'd be easier if everything were async, or there was an option to be sync or async.

Another option is that those integrations are `Posthog | Promise` and resolve the promises instead of bailing out.

benjackwhite:
I've thought about this a little before and as I see it there are two issues to solve:

1. Buffering capture type calls
2. Buffering get type calls


Capture calls are _somewhat_ straightforward as we would "simply" need to intercept the calling function and add it to a queue to be reapplied once the async initialisation is finished.

Get calls are trickier as we would need handle methods like getting the distinct id. Either we would return undefined if the sdk isn't initialised _or_ make all of these sort of methods async so that we can await the initialisation. 

I don't think these are so hard that we shouldn't do it, we just need to consider those cases and user expectations

marandaneto:
I think the user expectation would be that all events should be captured regardless the init is sync or async.

I'd also be sure that the timestamps are correct (the timestamp that was supposed to be captured) instead of the delayed timestamp, maybe a few milliseconds won't matter but worth considering, otherwise average session time might be affected due to delayed events.

PR #189, opened 2024-02-23
feat: Refactor the SDK to support a lazy initialization

benjackwhite:
## Problem
Attempt to resolve https://github.com/PostHog/posthog-js-lite/issues/188

The core SDK doesn't allow for later initialisation of the underlying storage. This moves a lot of complexity into the RN sub library. By adding support for a lazy initialisation the derived libraries can now "easily" have async loading of storage but provide a sync API to users which is much nicer to work with.

## Changes

* Wraps all internal methods with a check for the initPromise in a decorator-esque fashion which allows it to be sync or async
* Updates all the tests
* Removes initAsync in favour of the much simpler initialisation!
* Adds a `defaultOptIn` value in place of `enable`
* Adds a `disable` option for disabling the client fully (like for local dev testing)
* Swaps types to jsdoc style for autocomplete hints

**TODO**
- [ ]  Do we want to remove some deprecated methods / make feature flags fully async?

### Bump level

- [x] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [x] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

(Added to changelings)


benjackwhite:
@neilkakkar just pinging as this will require a major version so a good time to remove deprecated methods or change function calls if you had anything annoying in mind regarding feature flags (should be an additional PR but let me know before I actually merge this one and release it)

posthog-node/test/feature-flags.spec.ts
```diff
@@ -2051,81 +1987,74 @@ describe('match properties', () => {
...
-    const property_a = { key: 'key', value: '-6h', operator: 'is_date_before' }
-    expect(matchProperty(property_a, { key: '2022-03-01' })).toBe(true)
-    expect(matchProperty(property_a, { key: '2022-04-30' })).toBe(true)
-
+  it.each([
+    ['is_date_before', '-6h', '2022-03-01', false],
+    ['is_date_before', '-6h', '2022-04-30', true],
     // :TRICKY: MonthIndex is 0 indexed, so 3 is actually the 4th month, April.
-    expect(matchProperty(property_a, { key: new Date(Date.UTC(2022, 3, 30, 1, 2, 3)) })).toBe(true)
+    ['is_date_before', '-6h', new Date(Date.UTC(2022, 3, 30, 1, 2, 3)), true],
```
benjackwhite, line 1994: I had to do some refactoring here as the tests were really hard to debug and only seemed to start failing in this PR (still figuring out why)

posthog-node/test/feature-flags.spec.ts
```diff
@@ -13,71 +14,6 @@ const posthogImmediateResolveOptions: PostHogOptions = {
   fetchRetryCount: 0,
 }
 
-export const apiImplementation = ({
```
benjackwhite, line 16: Moved these out as another test imports them which essentially runs all the tests twice when you import one spec into another 🙈

posthog-core/src/types.ts
```diff
@@ -1,32 +1,34 @@
 export type PosthogCoreOptions = {
```
marandaneto, line 1: Since its a major version with breaking changes, should we rename types from `Posthog` to `PostHog`?

posthog-react-native/src/PostHogProvider.tsx
```diff
@@ -33,12 +33,6 @@ function PostHogLifecycleHook({ client }: { client?: PostHog }): JSX.Element | n
   return null
 }
 
-const isPromise = (value: any): value is Promise<PostHog> => {
```
marandaneto, line 36: CIAO hack `isPromise` :D

posthog-react-native/src/PosthogContext.ts
```diff
@@ -1,4 +1,4 @@
 import React from 'react'
 import { PostHog } from './posthog-rn'
 
-export const PostHogContext = React.createContext<{ client?: PostHog }>({ client: undefined })
+export const PostHogContext = React.createContext<{ client: PostHog }>({ client: undefined as unknown as PostHog })
```
marandaneto, line 4: Can we just remove `{ client: undefined as unknown as PostHog }`? since its declared as non nullable/undefined?

posthog-react-native/src/posthog-rn.ts
```diff
@@ -112,7 +66,13 @@ export class PostHog extends PostHogCore {
           this._semiAsyncStorage?.setItem(PostHogPersistedProperty.AnonymousId, legacyValues.anonymousId)
         }
       }
+    }
 
+    this._initPromise = setupAsync()
+
+    // Not everything needs to be in the init promise
+    this._initPromise.then(async () => {
+      this._isInitialized = true
```
marandaneto, line 75: This flag is being flipped but not being used anywhere, do we need it?
Either we always call `await this._initPromise` to be sure the promise is resolved or we check the `_isInitialized` flag, but right now seems not needed.

posthog-react-native/src/posthog-rn.ts
```diff
@@ -126,10 +86,14 @@ export class PostHog extends PostHogCore {
...
+
       await this.persistAppVersion()
-    }
+    })
+  }
 
-    this._setupPromise = setupAsync()
+  // NOTE: This is purely a helper method for testing purposes or those who wish to be certain the SDK is fully initialised
+  public async ready(): Promise<void> {
+    await this._initPromise
```
marandaneto, line 96: We could check `_isInitialized` first here.

marandaneto: Relates to https://github.com/PostHog/posthog-js-lite/pull/189/files#r1502255807

posthog-core/src/index.ts
```diff
@@ -55,16 +55,18 @@ export abstract class PostHogCoreStateless {
   private requestTimeout: number
   private captureMode: 'form' | 'json'
   private removeDebugCallback?: () => void
-  private debugMode: boolean = false
```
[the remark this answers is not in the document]
marandaneto: nvm its not used

posthog-react-native/src/posthog-rn.ts
marandaneto: nvm its used in the core project

posthog-core/src/index.ts
```diff
@@ -95,6 +97,24 @@ export abstract class PostHogCoreStateless {
...
     this.disableGeoip = options?.disableGeoip ?? true
+    this.disabled = options?.disabled ?? false
+    // Init promise allows the derived class to block calls until it is ready
+    this._initPromise = Promise.resolve()
+    this._isInitialized = true
+  }
+
+  protected wrap(fn: () => void): void {
+    if (this.disabled) {
+      // When disabled we don't do anything
```
marandaneto, line 108: Probably worth adding a console.log here (SDK is disabled).

posthog-core/src/types.ts
benjackwhite: good thinking - updated everywhere

posthog-react-native/src/PostHogProvider.tsx
benjackwhite: :D

posthog-react-native/src/PosthogContext.ts
benjackwhite: unfortunately not. The argument requires the type to match (which undefined doesn't). It's one of the quirks of context hooks sadly.

posthog-core/src/index.ts
```diff
@@ -696,7 +743,11 @@ export abstract class PostHogCore extends PostHogCoreStateless {
     }
   }
 
-  getSessionId(): string | undefined {
+  getSessionId(): string {
+    if (!this._isInitialized) {
+      return ''
```
marandaneto, line 748: Probably better to return undefined?
otherwise we need to add a check here https://github.com/PostHog/posthog-js-lite/blob/2d1da92df028f2915e5147c9fab07f77fd094f60/posthog-core/src/index.ts#L695
an empty string would trigger an ingestion warning I guess? it does with a few other important flags at least such as anon_id.

posthog-core/src/index.ts
```diff
@@ -709,10 +760,16 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
   resetSessionId(): void {
-    this.setPersistedProperty(PostHogPersistedProperty.SessionId, null)
+    this.wrap(() => {
+      this.setPersistedProperty(PostHogPersistedProperty.SessionId, null)
+    })
   }
 
   getAnonymousId(): string {
+    if (!this._isInitialized) {
+      return ''
```
marandaneto, line 770: Same as here https://github.com/PostHog/posthog-js-lite/pull/189/files#r1502270430

posthog-core/src/index.ts
```diff
@@ -709,10 +760,16 @@ export abstract class PostHogCore extends PostHogCoreStateless {
   }
 
   resetSessionId(): void {
```
marandaneto, line 762: Should it be async as well? since it needs to update the storage.

posthog-core/src/index.ts
benjackwhite: I thought about it but we'v had _a lot_ of people complain about our SDKs over-logging. Maybe I can make It only log if debug is on or something 🤔

posthog-core/src/index.ts
benjackwhite: Not sure tbh. I think the issue here is that if we make it undefined then the caller _always_ has to check for undefined. Is it nicer to force the user to always make that check or to just tolerate that it might be undefined.

Internal calls to that will never be undefined as they are always called from wrapped functions. Not sure what is more annoying for the user.

posthog-core/src/index.ts
benjackwhite: Not sure if you followed what the `wrap` code does, but essentially nothing is async in this function.

benjackwhite: It's just delayed execution _if_ necessary. If the SDK is initialised that function is entirely synchronous.

posthog-core/src/index.ts
marandaneto: Yeah, we need something similar to the other SDKs (android/ios).
A logging method that only prints to the console if `debug` is enabled.
https://github.com/PostHog/posthog-android/blob/main/posthog/src/main/java/com/posthog/internal/PostHogPrintLogger.kt
So we never call console.log directly but always our method that wraps `console.log`.
Not a blocker, I can also do this in a different PR another time.
I'll probably add a debug-level filter too.

posthog-core/src/index.ts
marandaneto: It's ok that it returns an empty (and invalid string) if written in the method docs, I also prefer that over `undefined`.

benjackwhite: I'll add some comments

marandaneto (approved):
Nicely done.

posthog-node/test/feature-flags.spec.ts
neilkakkar: Not sure if anything has changed here 🤔 , very happy to keep this refactor, but could you please first revert & check tests still pass, and then restore?

PR #191, opened 2024-02-26
fix: shutdownAsync flushes the whole queue

marandaneto:
## Problem

If the queue is bigger than `flushAt` after all `pendingPromises` are resolved, some events can still be lost since flush caps to `flushAt` anyway.
Discussed here with @benjackwhite https://github.com/PostHog/posthog-js-lite/pull/177#discussion_r1489359776

## Changes

Instead of calling `shutdownAsync` twice, we call until the queue is empty.

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [X] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X


posthog-core/src/index.ts
```diff
@@ -566,18 +566,26 @@ export abstract class PostHogCoreStateless {
...
             // ignore errors as we are shutting down and can't deal with them anyways.
           })
         )
       )
-      // flush again to make sure we send all events, some of which might've been added
-      // while we were waiting for the pending promises to resolve
-      // For example, see sendFeatureFlags in posthog-node/src/posthog-node.ts::capture
-      await this.flushAsync()
+
+      while (true) {
```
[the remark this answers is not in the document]
benjackwhite: Can we still have an escape hatch? I feel like we will for sure get some weird case where someone calls shutdown and ends up in a long while loop due to some weird configuration issue (like having flushat set to 1). 
30 second time out guard or something feels better than nothing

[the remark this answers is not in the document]
marandaneto: sure.

posthog-core/src/index.ts
```diff
@@ -616,23 +616,40 @@ export abstract class PostHogCoreStateless {
     )
   }
 
-  async shutdownAsync(): Promise<void> {
+  async shutdownAsync(shutdownTimeout?: number): Promise<void> {
```
benjackwhite, line 619: Just so its clear to the caller...
```suggestion
  async shutdownAsync(shutdownTimeoutMs?: number): Promise<void> {
```

posthog-core/src/index.ts
```diff
@@ -641,8 +658,8 @@ export abstract class PostHogCoreStateless {
     }
   }
 
-  shutdown(): void {
-    void this.shutdownAsync()
+  shutdown(shutdownTimeout?: number): void {
```
benjackwhite, line 661: Same here

benjackwhite (approved):
Other than a nit - love it!

package.json
```diff
@@ -43,14 +43,14 @@
     "eslint-plugin-react": "^7.30.1",
     "eslint-plugin-react-hooks": "^4.6.0",
     "husky": "^8.0.1",
-    "jest": "^28.1.3",
+    "jest": "^29.0.0",
```
marandaneto, line 46: Relates to https://github.com/jestjs/jest/issues/14741#issuecomment-1845219606 otherwise I cannot run tests locally

marandaneto:
@benjackwhite there's just one test failing but its unrelated to my changes I guess.

>  FAIL  posthog-node/test/feature-flags.spec.ts
>   ● match properties › with relative date operators: is_date_before, 1h, 2022-05-01 00:00:00
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: false
>     Received: true
> 
>       2040 |   ])('with relative date operators: %s, %s, %s', (operator, value, date, expectation) => {
>       2041 |     jest.setSystemTime(new Date('2022-05-01'))
>     > 2042 |     expect(matchProperty({ key: 'key', value, operator }, { key: date })).toBe(expectation)
>            |                                                                           ^
>       2043 |
>       2044 |     return
>       2045 |   })
> 
>       at posthog-node/test/feature-flags.spec.ts:2042:75
> 
>       

Could it be because of https://github.com/PostHog/posthog-js-lite/pull/189#discussion_r1502437416 ?

neilkakkar:
oh indeed, much better, thank you!

PR #196, opened 2024-02-29
chore: add retry default comments and fix misconfig default

marandaneto:
## Problem

There was a mismatch of the default retry timeout depending on the code path.

### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [X] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -91,9 +91,10 @@ export abstract class PostHogCoreStateless {
     // If enable is explicitly set to false we override the optout
     this.defaultOptIn = options?.defaultOptIn ?? true
 
+    // when changing the default, make sure to update the retriable method as well
     this._retryOptions = {
       retryCount: options?.fetchRetryCount ?? 3,
-      retryDelay: options?.fetchRetryDelay ?? 3000,
+      retryDelay: options?.fetchRetryDelay ?? 3000, // 3 seconds
       retryCheck: isPostHogFetchError,
     }
     this.requestTimeout = options?.requestTimeout ?? 10000 // 10 seconds
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -44,7 +44,7 @@ class PostHogFetchNetworkError extends Error {
 }
 
 function isPostHogFetchError(err: any): boolean {
-  return typeof err === 'object' && (err.name === 'PostHogFetchHttpError' || err.name === 'PostHogFetchNetworkError')
+  return typeof err === 'object' && (err instanceof PostHogFetchHttpError || err instanceof PostHogFetchNetworkError)
 }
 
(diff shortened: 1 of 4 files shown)


posthog-core/src/index.ts
```diff
@@ -581,11 +581,7 @@ export abstract class PostHogCoreStateless {
     })
   }
 
-  private async fetchWithRetry(
-    url: string,
-    options: PostHogFetchOptions,
-    retryOptions?: RetriableOptions
-  ): Promise<PostHogFetchResponse> {
+  private async fetchWithRetry(url: string, options: PostHogFetchOptions): Promise<PostHogFetchResponse> {
```
marandaneto, line 584: `retryOptions` was never used here.

posthog-core/src/utils.ts
```diff
@@ -9,27 +9,26 @@ export function removeTrailingSlash(url: string): string {
...
-  retryCount?: number
-  retryDelay?: number
-  retryCheck?: (err: any) => boolean
+  retryCount: number
+  retryDelay: number
+  retryCheck: (err: any) => boolean
 }
 
-export async function retriable<T>(fn: () => Promise<T>, props: RetriableOptions = {}): Promise<T> {
-  const { retryCount = 3, retryDelay = 5000, retryCheck = () => true } = props
```
marandaneto, line 18: there was a mismatch of the default value of `retryDelay`, either 3s or 5s depending on the code path, refactored to make it non-optional since `_retryOptions` is always created in the ctor with default values anyway.

posthog-core/src/index.ts
```diff
@@ -44,7 +44,7 @@ class PostHogFetchNetworkError extends Error {
 }
 
 function isPostHogFetchError(err: any): boolean {
-  return typeof err === 'object' && (err.name === 'PostHogFetchHttpError' || err.name === 'PostHogFetchNetworkError')
+  return typeof err === 'object' && (err instanceof PostHogFetchHttpError || err instanceof PostHogFetchNetworkError)
```
marandaneto, line 47: @benjackwhite was there a reason the checks were not typesafe instead?

marandaneto: the checks could fail if the code is aggressively minified

benjackwhite: This was an external contribution iirc so 🤷

marandaneto: true https://github.com/PostHog/posthog-js-lite/pull/88/

neilkakkar (approved):

neilkakkar:
Will follow up with custom defaults for flags as well 👍

Issue #195, opened 2024-02-28
Fetching feature flags retries for too long - not configurable


paularmstrong:
## Bug description

There is currently an outage on the posthog service side for feature flags. This is causing my server to thrash – continually logging network errors when trying to call `posthog.getAllFlags()`

This is caused by `fetchWithRetry`, which by default calls 4 times, with a 5 second delay between each call, resulting in requests on the server stalling for 20+ seconds: https://github.com/PostHog/posthog-js-lite/blob/c59c9d6b13bc972e0b1b3b66fe7a6c6a0e22f0ab/posthog-core/src/index.ts#L295

This retry is not configurable from the `getAllFlags()` endpoint (or any others that I'm seeing)

## How to reproduce

1. Take posthog down
2. Call `posthog.getAllFlags()`
3. Wait forever

## Related sub-libraries

- [ ] All of them
- [x] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

PR #197, opened 2024-02-29
fix(flags): Add a shorter timeout, disable retries for flags

neilkakkar:
## Problem

^^ We want a much shorter timeout on flags, since they can block client applications. Also makes this configurable now.

Fixes https://github.com/PostHog/posthog-js-lite/issues/195

### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -54,6 +54,7 @@ export abstract class PostHogCoreStateless {
   private flushAt: number
   private flushInterval: number
   private requestTimeout: number
+  private featureFlagsRequestTimeout: number
   private captureMode: 'form' | 'json'
   private removeDebugCallback?: () => void
   private disableGeoip: boolean = true
@@ -97,6 +98,7 @@ export abstract class PostHogCoreStateless {
       retryCheck: isPostHogFetchError,
     }
     this.requestTimeout = options?.requestTimeout ?? 10000 // 10 seconds
+    this.featureFlagsRequestTimeout = options?.featureFlagsRequestTimeout ?? 3000 // 3 seconds
     this.disableGeoip = options?.disableGeoip ?? true
     this.disabled = options?.disabled ?? false
     // Init promise allows the derived class to block calls until it is ready
@@ -292,7 +294,8 @@ export abstract class PostHogCoreStateless {
         ...extraPayload,
       }),
     }
-    return this.fetchWithRetry(url, fetchOptions)
+    // Don't retry /decide API calls
+    return this.fetchWithRetry(url, fetchOptions, { retryCount: 0 }, this.featureFlagsRequestTimeout)
       .then((response) => response.json() as Promise<PostHogDecideResponse>)
       .catch((error) => {
         this._events.emit('error', error)
@@ -581,7 +584,7 @@ export abstract class PostHogCoreStateless {
     })
   }
 
-  private async fetchWithRetry(url: string, options: PostHogFetchOptions): Promise<PostHogFetchResponse> {
+  private async fetchWithRetry(url: string, options: PostHogFetchOptions, retryOptions?: Partial<RetriableOptions>, requestTimeout?: number): Promise<PostHogFetchResponse> {
     ;(AbortSignal as any).timeout ??= function timeout(ms: number) {
       const ctrl = new AbortController()
       setTimeout(() => ctrl.abort(), ms)
@@ -593,7 +596,7 @@ export abstract class PostHogCoreStateless {
         let res: PostHogFetchResponse | null = null
         try {
           res = await this.fetch(url, {
-            signal: (AbortSignal as any).timeout(this.requestTimeout),
+            signal: (AbortSignal as any).timeout(requestTimeout ?? this.requestTimeout),
             ...options,
           })
         } catch (e) {
@@ -609,7 +612,7 @@ export abstract class PostHogCoreStateless {
         }
         return res
       },
-      { ...this._retryOptions }
+      { ...this._retryOptions, ...retryOptions }
     )
   }
 
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -54,7 +54,7 @@ export abstract class PostHogCoreStateless {
   private flushAt: number
   private flushInterval: number
   private requestTimeout: number
-  private featureFlagsRequestTimeout: number
+  private featureFlagsRequestTimeoutMs: number
   private captureMode: 'form' | 'json'
   private removeDebugCallback?: () => void
   private disableGeoip: boolean = true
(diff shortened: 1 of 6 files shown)


posthog-core/src/index.ts
```diff
@@ -292,7 +294,8 @@ export abstract class PostHogCoreStateless {
         ...extraPayload,
       }),
     }
-    return this.fetchWithRetry(url, fetchOptions)
+    // Don't retry /decide API calls
+    return this.fetchWithRetry(url, fetchOptions, { retryCount: 0 }, this.featureFlagsRequestTimeout)
```
benjackwhite, line 298: If we have retrycount 0 then do we even want to call `fetchWithRetry` anymore? 🤔

posthog-node/src/posthog-node.ts
```diff
@@ -114,12 +113,20 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
       super.captureStateless(distinctId, event, props, { timestamp, disableGeoip, uuid })
     }
 
+    const _getFlags = (
```
benjackwhite, line 116: Why this abstraction? Feels like a lot more code that doesn't do anything?

benjackwhite (approved):
Nitty comments but otherwise 👍

posthog-node/src/posthog-node.ts
neilkakkar: You were too fast with the review lol was just writing a letter to you xD

posthog-core/src/index.ts
neilkakkar: Good question, `fetchWithRetry` also handles timeouts, and didn't want to introduce yet another function just for timeouts. I could rename it to `fetchWithRetryAndTimeout` ?

posthog-core/src/index.ts
```diff
@@ -97,6 +98,7 @@ export abstract class PostHogCoreStateless {
       retryCheck: isPostHogFetchError,
     }
     this.requestTimeout = options?.requestTimeout ?? 10000 // 10 seconds
+    this.featureFlagsRequestTimeout = options?.featureFlagsRequestTimeout ?? 3000 // 3 seconds
```
marandaneto, line 101: 3s might be very low for emerging countries, I know it's configurable but wondering if it's a good default, maybe 5s? thinking of Mobile specifically where the internet is much worse.

posthog-core/src/index.ts
marandaneto: `fetchWithRetryAndOrTimeout` :D

marandaneto:
@neilkakkar the original issue is about `posthog.getAllFlags()` which forwards the call to `getAllFlagsAndPayloads`, then `_loadFeatureFlags`.
`_loadFeatureFlags` seems to have its own timeout which is 10s, should the `api/feature_flag` endpoint share the decide timeout or is it something else?

marandaneto:
I'm wondering how the signal abort timeout is computed, didn't find anything here https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static
The timeout must be just idle IMO, instead of the whole thing, for example, connect, DNS handshake, waiting for API request, downloading bytes, answering, etc, 3s would be very low for slow network/emerging countries as a default.

posthog-core/src/index.ts
marandaneto: More here https://github.com/PostHog/posthog-js-lite/pull/197#issuecomment-1972690919

neilkakkar:
The `loadFeatureFlags` timeout is for local evaluation flags which are cached, so can have a much longer timeout. I don't think that's where the problem came in here, it's after, when you call `decide/` + retry.

The default timeout for all requests works for local eval as well, since generally it needs to succeed only once, rather than everytime we call `getFeatureFlag` or `getAllFlags()`

posthog-core/src/index.ts
neilkakkar: I'm basing the timeout on the request latency + external monitor latency: https://status.posthog.com/

But you're right the external monitor probably is also close by.

Although, node should be on a server, maybe we can have a different larger default for react-native? My worry is it might be too long on any server if it blocks for 5s by default.

marandaneto: That's why some HTTP clients have more timeout configurations, such as connect (which involves DNS resolving, SSL handshaking, etc), read, write, etc.

Apparently, here the abort signal is all or nothing which is suboptimal.

Let's keep it 10s for RN, I already had issues with 5s default with Mobile SDKs, and 10s was the magic number that worked better for emerging countries or very slow/spotty connections.

neilkakkar: Reasonable, will keep the default for RN and let people customise it 👍

marandaneto:
@neilkakkar once you address https://github.com/PostHog/posthog-js-lite/pull/197#discussion_r1510286304 we can merge and cut the first preview of the new major version of the SDKs

posthog-core/src/types.ts
```diff
@@ -26,6 +26,8 @@ export type PostHogCoreOptions = {
   fetchRetryDelay?: number
   /** Timeout in milliseconds for any calls. Defaults to 10 seconds. */
   requestTimeout?: number
+  /** Timeout in milliseconds for feature flag calls. Defaults to 3 seconds. */
+  featureFlagsRequestTimeoutMs?: number
```
marandaneto, line 30: @neilkakkar the comment here should mention that for stateful is 10s

posthog-node/src/posthog-node.ts
```diff
@@ -114,12 +113,20 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
...
+      return super.getFeatureFlagsStateless(distinctId, groups, undefined, undefined, disableGeoip)
+    }
+
     // :TRICKY: If we flush, or need to shut down, to not lose events we want this promise to resolve before we flush
     const capturePromise = Promise.resolve()
       .then(async () => {
         if (sendFeatureFlags) {
           // If we are sending feature flags, we need to make sure we have the latest flags
-          return await super.getFeatureFlagsStateless(distinctId, groups, undefined, undefined, disableGeoip)
+          return await _getFlags(distinctId, groups, disableGeoip)
```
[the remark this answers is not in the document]
marandaneto: @neilkakkar is this resolved?

marandaneto (approved):
Left one comment and one question, otherwise LGTM.

posthog-core/src/types.ts
neilkakkar: oof nice catch thanks!

Issue #96, opened 2023-05-02
Improve/add support for sync providers, ie rn-mmkv


filipef101:
## Is your feature request related to a problem?

Would be nice if we could use MMKV instead of async storage

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

filipef101:
Some people may have mmkv instead of async storage, posthog could detect if async storage not present fallback to mmkv + option to prefer mmkv

benjackwhite:
We didn't want to end up with 100s of if statements trying to cover every possible dependency 😅 
But you can just configure your own interface like so:


```ts
const options: PostHogOptions = {
  customAsyncStorage: {
    getItem: (key: string) => Promise<string | null>,
    setItem: (key: string, value: string) => Promise<void>
  }
}
```

filipef101:
mmkv is sync, any chance to get a sync storage provider? to be able to have feature flags on first render/paint @benjackwhite

benjackwhite:
I'm not 100% certain but I would expect if you simply resolve immediately in those custom functions then it basically will be sync, especially from a first-render point of view?

marandaneto:
The bridge is async by design, so even if you resolve/await a function, it does not guarantee the order of execution, it might be that something is executed before your awaited function (it's an event loop/single thread), that has also been awaited for.

A solution for all the async problems is to migrate to the [RN new architecture](https://reactnative.dev/docs/next/the-new-architecture/why#new-architectures-improvements) and define when a method is async/sync or even have both.
Since all the storage plugins are Native libs (android/ios), this would be the best/most reliable option.

For most cases, just awaiting the methods is enough, showing loading progress. At the same time, the calls get resolved, and after you decide what to do depending on the resolved feature flags, it's not necessary to be before the first render/paint, you can work around that by just using this suggestion for now.

filipef101:
@marandaneto I don't see any native code, so what needs to be migrated or am i missing something?

filipef101:
> I'm not 100% certain but I would expect if you simply resolve immediately in those custom functions then it basically will be sync, especially from a first-render point of view?

@benjackwhite doesnt matter how fast it is, if you need to await it wont be available on first paint/render

marandaneto:
> @marandaneto I don't see any native code, so what needs to be migrated or am i missing something?

[...]

PR #198, opened 2024-03-01
chore: ref to allow sync storage friendlier

marandaneto:
## Problem

Tempt to fix https://github.com/PostHog/posthog-js-lite/issues/96

### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/src/posthog-rn.ts
+++ b/posthog-react-native/src/posthog-rn.ts
@@ -10,11 +10,12 @@ import {
 } from '../../posthog-core/src'
 import { PostHogMemoryStorage } from '../../posthog-core/src/storage-memory'
 import { getLegacyValues } from './legacy'
-import { SemiAsyncStorage } from './storage'
+import { SemiAsyncStorage, Storage, SyncStorage } from './storage'
 import { version } from './version'
 import { buildOptimisiticAsyncStorage, getAppProperties } from './native-deps'
-import { PostHogAutocaptureOptions, PostHogCustomAppProperties, PostHogCustomAsyncStorage } from './types'
+import { PostHogAutocaptureOptions, PostHogCustomAppProperties, PostHogCustomAsyncStorage, PostHogCustomSyncStorage } from './types'
 import { withReactNativeNavigation } from './frameworks/wix-navigation'
+import AsyncStorage from '@react-native-async-storage/async-storage'
 
 export type PostHogOptions = PostHogCoreOptions & {
   persistence?: 'memory' | 'file'
@@ -22,14 +23,14 @@ export type PostHogOptions = PostHogCoreOptions & {
   customAppProperties?:
     | PostHogCustomAppProperties
     | ((properties: PostHogCustomAppProperties) => PostHogCustomAppProperties)
-  customAsyncStorage?: PostHogCustomAsyncStorage
+  customStorage?: PostHogCustomAsyncStorage | PostHogCustomSyncStorage
   captureNativeAppLifecycleEvents?: boolean
 }
 
 export class PostHog extends PostHogCore {
   private _persistence: PostHogOptions['persistence']
   private _memoryStorage = new PostHogMemoryStorage()
-  private _semiAsyncStorage?: SemiAsyncStorage
+  private _storage?: Storage
   private _appProperties: PostHogCustomAppProperties = {}
 
   constructor(apiKey: string, options?: PostHogOptions) {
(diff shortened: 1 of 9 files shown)


posthog-react-native/src/posthog-rn.ts
```diff
@@ -48,22 +49,31 @@ export class PostHog extends PostHogCore {
...
-      this._semiAsyncStorage = new SemiAsyncStorage(options?.customAsyncStorage || buildOptimisiticAsyncStorage())
+      const storage = options?.customStorage || buildOptimisiticAsyncStorage()
+
+      if (storage.isSemiAsync()) {
+        const asyncStorage = new SemiAsyncStorage(storage as PostHogCustomAsyncStorage)
+        this._storage = asyncStorage
+      } else {
+        const syncStorage = new SyncStorage(storage as PostHogCustomSyncStorage)
+        this._storage = syncStorage
+        syncStorage.preload()
```
marandaneto, line 60: In this case, it loads sync.

posthog-react-native/src/posthog-rn.ts
```diff
@@ -48,22 +49,31 @@ export class PostHog extends PostHogCore {
...
+      }
     }
 
     // Ensure the async storage has been preloaded (this call is cached)
     const setupAsync = async (): Promise<void> => {
-      if (!this._semiAsyncStorage?.isPreloaded) {
-        await this._semiAsyncStorage?.preloadAsync()
+      if (!this._storage?.isPreloaded && this._storage?.isSemiAsync()) {
+        await (this._storage as SemiAsyncStorage)?.preloadAsync()
```
marandaneto, line 67: Loads async if async

posthog-react-native/src/posthog-rn.ts
```diff
@@ -48,22 +53,33 @@ export class PostHog extends PostHogCore {
...
+      const storage = options?.customStorage || buildOptimisiticAsyncStorage()
+
+      if (storage.isSemiAsync()) {
+        const asyncStorage = new SemiAsyncStorage(storage as PostHogCustomAsyncStorage)
+        this._storage = asyncStorage
+      } else {
+        const syncStorage = new SyncStorage(storage as PostHogCustomSyncStorage)
+        this._storage = syncStorage
+        if (!syncStorage.isPreloaded) {
+          syncStorage.preload()
```
marandaneto, line 65: loads sync if sync

marandaneto:
Definitely breaking changes but in good time if the approach is better for sync storages.
cc @filipef101 WDYT?

posthog-react-native/src/posthog-rn.ts
```diff
@@ -66,9 +70,7 @@ export class PostHog extends PostHogCore {
 
     // Ensure the async storage has been preloaded (this call is cached)
     const setupAsync = async (): Promise<void> => {
-      if (!this._asyncStorage?.isPreloaded) {
-        await (this._asyncStorage as SemiAsyncStorage)?.preloadAsync()
-      }
+      await (this._asyncStorage as SemiAsyncStorage)?.preloadAsync()
```
marandaneto, line 73: `preloadAsync` will check `isPreloaded` already

posthog-react-native/src/posthog-rn.ts
```diff
@@ -22,14 +26,19 @@ export type PostHogOptions = PostHogCoreOptions & {
   customAppProperties?:
     | PostHogCustomAppProperties
     | ((properties: PostHogCustomAppProperties) => PostHogCustomAppProperties)
+  /** Allows you to provide a custom and synchronous storage such as mmkv */
+  customStorage?: PostHogCustomSyncStorage
+  /** Allows you to provide a custom and asynchronous storage such as async-storage, expo-file-system or something else.
+   * By default, the SDK will try to load expo-file-system and then async-storage if none is provided.
+   */
```
marandaneto, line 33: @benjackwhite the downside of having both types is that if a user passes both, `customStorage` will have precedence which is fine but it's unclear for the user, I will add as a comment

posthog-react-native/src/storage.ts
```diff
@@ -1,19 +1,75 @@
...
+  populateMemoryCache(res: string | null): void {
+    try {
+      const data = res ? JSON.parse(res).content : {}
+
+      for (const key in data) {
+        this.memoryCache[key] = data[key]
+      }
+    } catch (e) {
+      console.warn(
+        "PostHog failed to load persisted data from storage. This is likely because the storage format is. We'll reset the storage.",
```
marandaneto, line 56: @benjackwhite I dont think we reset the storage though, the comment is outdated?

posthog-react-native/src/posthog-rn.ts
benjackwhite: Got rid of it alltogether

posthog-react-native/src/posthog-rn.ts
benjackwhite: Added a warning as well if both are set.

posthog-react-native/src/posthog-rn.ts
```diff
@@ -47,31 +71,43 @@ export class PostHog extends PostHogCore {
...
 
-    // Ensure the async storage has been preloaded (this call is cached)
-    const setupAsync = async (): Promise<void> => {
-      if (!this._semiAsyncStorage?.isPreloaded) {
-        await this._semiAsyncStorage?.preloadAsync()
+      if (options?.customStorage) {
+        this._storage = new PostHogRNSyncStorage(options.customStorage)
+      } else {
+        this._storage = new PostHogRNSemiAsyncStorage(options?.customAsyncStorage || buildOptimisiticAsyncStorage())
+        storagePromise = this._storage.preloadPromise
```
filipef101, line 87: I would just treat `customStorage` and `customAsyncStorage` the same, and deprecate `customAsyncStorage`. Calling preloadPromise if exists.
`customStorage` typed as generic `PostHogRNStorage` or whatever is generic.

posthog-react-native/src/types.ts
```diff
@@ -51,3 +51,8 @@ export interface PostHogCustomAsyncStorage {
   getItem: (key: string) => Promise<string | null>
   setItem: (key: string, value: string) => Promise<void>
 }
+
+export interface PostHogCustomSyncStorage {
+  getItem: (key: string) => string | null
+  setItem: (key: string, value: string) => void
+}
```
filipef101, line 58: Not redundant?

I mean, maybe the type of the getItem on the above `PostHogCustomAsyncStorage` is wrong, and maybe should be a single one with an added optional method  `preloadPromise`.

Then both sync and async storages classes can just implement this storage, instead of extending or whatever

filipef101 (commented):
I'm happy with this, and think fits my needs. Might be nitpicking or confused, the whole async vs non async storage feels irrelevant. Think should be a single type/usage etc. where an optional `preloadPromise` can be implemented (as implemented on default storage).  Not sure if that makes sense. I did something like that on my proposal PR for context https://github.com/PostHog/posthog-js-lite/pull/170 

I don't care much really about changing anything, just some thoughts.

image

posthog-react-native/test/storage.spec.ts
```diff
@@ -23,14 +22,13 @@ describe('PostHog React Native', () => {
...
       })
+
+      storage = new PostHogRNSemiAsyncStorage(buildOptimisiticAsyncStorage())
     })
 
     it('should load storage from the file system', async () => {
       expect(storage.getItem('foo')).toEqual(undefined)
-      storage.preloadAsync()
-      storage.preloadAsync()
-      storage.preloadAsync()
```
marandaneto, line 32: Nice removal =)

posthog-react-native/src/posthog-rn.ts
benjackwhite: Good call.

benjackwhite (approved):
I think this is good to go

marandaneto:
> I think this is good to go

Teamwork and the community jumping in cc @filipef101 
Thanks!

PR #199, opened 2024-03-04
chore: release next major beta-1

marandaneto:
### Bump level

- [X] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [X] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-node/CHANGELOG.md
+++ b/posthog-node/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Next (4.0.0)
+# 4.0.0-beta.1 - 2024-03-04
 
 - Adds a `disabled` option and the ability to change it later via `posthog.disabled = true`. Useful for disabling PostHog tracking for example in a testing environment without having complex conditional checking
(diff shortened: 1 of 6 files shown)


benjackwhite (approved):
If it had been me I wouldn't have bothered with a beta release.

FYI - the current deploy action will actually mark this as an alpha release in npm. It defaults to `alpha` as the tag and only uses `latest` if the semver is a simple `a.b.c`

marandaneto:
> If it had been me I wouldn't have bothered with a beta release.
> 
> FYI - the current deploy action will actually mark this as an alpha release in npm. It defaults to `alpha` as the tag and only uses `latest` if the semver is a simple `a.b.c`

I think giving a few days after some structural changes is always good, already saved me before, and also give time to update the docs, etc.
No one is depending on this release as well, let's play it safe.

PR #200, opened 2024-03-05
feat: add uiHost

raquelmsmith:
## Problem

We want to reverse proxy our traffic to our flag and capture endpoints so we don't run into issues due to ad blockers. 

Reverse proxy docs say to add ui_host to the init. I think this is where I do it... I haven't committed here before and don't really know how this repo works so I'm mostly just feeling around in the dark here 😅 

## Changes

Adds `uiHost` in a variety of places. Pairs with https://github.com/PostHog/posthog/pull/20698

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [x] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -51,6 +51,7 @@ export abstract class PostHogCoreStateless {
   // options
   private apiKey: string
   host: string
+  uiHost?: string
   private flushAt: number
   private flushInterval: number
   private requestTimeout: number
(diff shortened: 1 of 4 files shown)


benjackwhite:
This doesn't make any sense here. It only applies for generating URLs to the web app (which js-lite doesn't do)

PR #206, opened 2024-03-13
fix: PostHogProvider initialization logic

AdamDorwart:
## Problem

A regression introduced by #189 requires a client and apiKey to initialize a PostHogProvider even though the error message says only one is required.

## Changes

The error condition logic is fixed and a non null assertion is made.

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Fix: PostHogProvider initialization

### DIFF
--- a/posthog-react-native/CHANGELOG.md
+++ b/posthog-react-native/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Next
 
 1. Sets `User-Agent` headers with SDK name and version for RN
+2. fix: PostHogProvider initialization that requires client `or` apiKey and not `and`.
 
 # 3.0.0-beta.2 - 2024-03-12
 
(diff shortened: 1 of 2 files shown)


benjackwhite (approved):
oooh crikey that is definitely an oversight. Thanks for finding that!

AdamDorwart:
I'm pretty fresh to Typescript. How would you prefer I resolve:
```
Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
     return client ?? new PostHog(apiKey!, options)
```
I could just
```
    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```

This also works but I'm not sure if there's a more idiomatic way to go about it without the throw at the end.
```typescript
    if (client) {
      return client
    } else if (apiKey) {
      return new PostHog(apiKey, options) 
    }

    throw new Error('Type logic error: This should never happen')
```

marandaneto:
@AdamDorwart I fixed the linting issue by just adding `apiKey ?? ''` since this code path won't be triggered anyway, less code and easier to read, also added the changelog.
Thanks!

PR #208, opened 2024-03-14
chore: add maxQueueSize support

marandaneto:
## Problem

One more improvement for https://github.com/PostHog/posthog-js-lite/issues/176

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [X] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -90,7 +90,7 @@ export abstract class PostHogCoreStateless {
     this.host = removeTrailingSlash(options?.host || 'https://app.posthog.com')
     this.flushAt = options?.flushAt ? Math.max(options?.flushAt, 1) : 20
     this.maxBatchSize = Math.max(this.flushAt, options?.maxBatchSize ?? 100)
-    this.maxQueueSize = options?.maxQueueSize ?? 1000
+    this.maxQueueSize = Math.max(this.flushAt, options?.maxQueueSize ?? 1000)
     this.flushInterval = options?.flushInterval ?? 10000
     this.captureMode = options?.captureMode || 'form'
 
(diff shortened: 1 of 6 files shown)


marandaneto:
@benjackwhite we forgot about this, otherwise, the storage might blow people's devices since events are cached now, either in memory or locally.
will add test it, just confirming if there was not one yet for some reason.

posthog-core/src/index.ts
```diff
@@ -89,6 +90,7 @@ export abstract class PostHogCoreStateless {
     this.host = removeTrailingSlash(options?.host || 'https://app.posthog.com')
     this.flushAt = options?.flushAt ? Math.max(options?.flushAt, 1) : 20
     this.maxBatchSize = Math.max(this.flushAt, options?.maxBatchSize ?? 100)
+    this.maxQueueSize = options?.maxQueueSize ?? 1000
```
pauldambra, line 93: should we log or throw here if maxQueueSize <= flushAt ??

i'd be tempted to have an even higher limit to reduce risk of dropping events but i guess we don't know what a reasonable number is 🤷

pauldambra (approved):
👍

posthog-core/src/index.ts
marandaneto: All SDKs default to 1k right now, at least the ones I checked.
`maxBatchSize` also does not log out nor throw.
it's not that you `must` otherwise nothing works, it's just that does not make much sense.
when the timer that flushes periodically is triggered, it does not care if `flushAt` is reached yet and will flush all pending events anyway.

marandaneto: I can do the `Math.max` though.

PR #209, opened 2024-03-18
chore: release next major GA

marandaneto:
## Problem

* Cut GA version
* Split contributing from readme
* Split changelog from readme

### Bump level

- [X] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [X] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X


posthog-node/CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
...
+1. Adds a `disabled` option and the ability to change it later via `posthog.disabled = true`. Useful for disabling PostHog tracking for example in a testing environment without having complex conditional checking
+2. Fixes some typos in types
+3. `shutdown` and `shutdownAsync` takes a `shutdownTimeoutMs` param with a default of 30000 (30s). This is the time to wait for flushing events before shutting down the client. If the timeout is reached, the client will be shut down regardless of pending events.
+4. Adds a new `featureFlagsRequestTimeoutMs` timeout parameter for feature flags which defaults to 3 seconds, updated from the default 10s for all other API calls.
+5. `flushAsync` and `shutdownAsync` are removed with `flush` and `shutdown` now being the async methods.
+6. Fixed an issue where `shutdown` would potentially exit early if a flush was already in progress
+7. Flushes will now try to flush up to `maxBatchSize` (default 100) in one go
+8. Sets `User-Agent` headers with SDK name and version for RN
+9. Queued events are limited up to `maxQueueSize` (default 1000) and the oldest events are dropped when the limit is reached
```
benjackwhite, line 11:
```suggestion
1. Adds a `disabled` option and the ability to change it later via `posthog.disabled = true`. Useful for disabling PostHog tracking for example in a testing environment without having complex conditional checking
2. `flushAsync` and `shutdownAsync` are removed with `flush` and `shutdown` now being the async methods.
3. `shutdown` takes a `shutdownTimeoutMs` param with a default of 30000 (30s). This is the time to wait for flushing events before shutting down the client. If the timeout is reached, the client will be shut down regardless of pending events.
4. Flushes will now try to flush up to `maxBatchSize` (default 100) in one go
5. Queued events are limited up to `maxQueueSize` (default 1000) and the oldest events are dropped when the limit is reached
6. Adds a new `featureFlagsRequestTimeoutMs` timeout parameter for feature flags which defaults to 3 seconds, updated from the default 10s for all other API calls.
7. Fixed an issue where `shutdown` would potentially exit early if a flush was already in progress
8. Sets `User-Agent` headers with SDK name and version for RN
9. Fixes some typos in types
```

benjackwhite (changes requested):
Suggested changed order to the changelog (only checked out one of them) so it is more readable and doesn't mention things like `shutdownAsync` given it is removed.

Issue #144, opened 2023-12-19
`Application Became Active` event on RN should have been `Application Opened` with `from_background=true` property


marandaneto:
### Description

* [Code](https://github.com/PostHog/posthog-js-lite/blob/4b24090d063e8b9472860f1ad0af7c1c6a21b4d8/posthog-react-native/src/hooks/useLifecycleTracker.ts#L16-L25) and [here](https://github.com/PostHog/posthog-js-lite/blob/01fff0242a620345614a792379397029cf6cc9f1/posthog-react-native/src/posthog-rn.ts#L239-L250).
* [Docs](https://posthog.com/docs/libraries/react-native#autocapture)

RN has a specific event called `Application Became Active` when the app is [active](https://reactnative.dev/docs/appstate#app-states) which means the app is back in the foreground.
On the other Mobile SDKs, we just capture the `Application Opened` event with the `from_background=true` property instead.

Ideally, we'd align this behavior in one way or the other.

When mobile dashboard templates are shipped, templates should have a single source of truth for events and properties.

benjackwhite:
I think at the time there was a back and forth with customers and we agreed having a dedicated name was better (with a view to eventually moving the iOS and Android over with a breaking major version change). But that was a long time ago so my memory may be hazy

marandaneto:
Yeah, I imagine filtering events/creating insights would be easier with its active event.
Can consider it for the next major though.

marandaneto:
We'll address that on the Android/iOS version instead in the next major.

PR #211, opened 2024-03-25
chore: Change Application opened to application became active for native events

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js-lite/issues/144
https://posthoghelp.zendesk.com/agent/tickets/11961
Depend on the SDK config, the dashboard should use different event names, its better to align that within the SDK

## Changes

Technically a breaking change but I don't see many people using `captureNativeAppLifecycleEvents` anyways, it's written in the changelog.
It's hard for support to figure out the multiple scenarios depending on the SDK configuration.
Our mobile dashboard templates don't even consider the `from_background` property anyway, so now using this config (captureNativeAppLifecycleEvents) will match our templates.
Added the app version and build to the other integrations.

### Bump level

- [ ] Major
- [X] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/src/frameworks/wix-navigation.ts
+++ b/posthog-react-native/src/frameworks/wix-navigation.ts
@@ -23,12 +23,22 @@ export const withReactNativeNavigation = (posthog: PostHog, options: PostHogAuto
       return
     }
 
-    posthog.capture('Application Opened')
+    const appProperties = posthog.getAppProperties()
+    const appBuild = appProperties.$app_build
+    const appVersion = appProperties.$app_version
+
+    posthog.capture('Application Opened', {
+      version: appVersion,
+      build: appBuild,
+    })
 
     AppState.addEventListener('change', (nextAppState) => {
       switch (nextAppState) {
         case 'active':
-          return posthog.capture('Application Became Active')
+          return posthog.capture('Application Became Active', {
+            version: appVersion,
+            build: appBuild,
+          })
         case 'background':
           return posthog.capture('Application Backgrounded')
         default:
--- a/posthog-react-native/src/frameworks/wix-navigation.ts
+++ b/posthog-react-native/src/frameworks/wix-navigation.ts
@@ -27,6 +27,7 @@ export const withReactNativeNavigation = (posthog: PostHog, options: PostHogAuto
     const appBuild = appProperties.$app_build
     const appVersion = appProperties.$app_version
 
+    // TODO: add missing initialUrl
     posthog.capture('Application Opened', {
       version: appVersion,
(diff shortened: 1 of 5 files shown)


posthog-react-native/src/frameworks/wix-navigation.ts
```diff
@@ -23,12 +23,23 @@ export const withReactNativeNavigation = (posthog: PostHog, options: PostHogAuto
       return
     }
 
-    posthog.capture('Application Opened')
+    const appProperties = posthog.getAppProperties()
+    const appBuild = appProperties.$app_build
+    const appVersion = appProperties.$app_version
+
+    // TODO: add missing initialUrl
+    posthog.capture('Application Opened', {
```
marandaneto, line 31: All events already contain `$app_build` and `$app_version` but for some reason (maybe legacy props), all the other SDKs were sending `version` and `build` as well for the lifecycle events.

marandaneto: An option is to remove `version` and `build` from all those events as well, the mobile dashboards just rely on `$app_version` anyway, not sure if there's any other feature I am unaware of.

benjackwhite (changes requested):

posthog-react-native/src/hooks/useLifecycleTracker.ts
```diff
@@ -9,14 +9,25 @@ export function useLifecycleTracker(client?: PostHog): void {
...
+      })
     }
     const subscription = AppState.addEventListener('change', (nextAppState) => {
       switch (nextAppState) {
         case 'active':
-          return posthog.capture('Application Became Active')
+          return posthog.capture('Application Became Active', {
+            version: appVersion,
+            build: appBuild,
+          })
```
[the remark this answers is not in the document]
benjackwhite: Eugh. Personally I'd much rather align around $ properties than this change as it feels really un-posthog.

Would it be possible to do the opposite? Change the other SDKs to include the $ versions?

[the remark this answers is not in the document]
marandaneto: removed them, will remove in the next major, and added a deprecated comment

posthog-react-native/src/frameworks/wix-navigation.ts
marandaneto: Will remove them in the next major https://github.com/PostHog/posthog-js-lite/issues/212

PR #219, opened 2024-04-24
fix(flags): Handle payloads correctly

neilkakkar:
## Problem

Fixes https://github.com/PostHog/posthog/issues/21768

Noticed payloads are handled very messily, and some existing tests for these were wrong 🫠. Now always simulates how things should be in production, with the decide response returning stringified payloads.

The code now always parses payloads once: when we get the response from decide (or local eval), and everything else can assume the payload is an object. This means we don't need to faff around parsing everywhere, which is nice.

Ran locally against production and can confirm everything works nicely now.

### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-node/CHANGELOG.md
+++ b/posthog-node/CHANGELOG.md
@@ -1,3 +1,8 @@
+
+# 4.0.1 - 2024-04-24
+
+1. Prevent double JSON parsing of feature flag payloads, which would convert the payload [1] into 1.
+
 # 4.0.0 - 2024-03-18
 
 ## Added
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -148,11 +148,15 @@ class FeatureFlagsPoller {
     }
 
     // Undefined means a loading or missing data issue. Null means evaluation happened and there was no match
-    if (response === undefined) {
+    if (response === undefined || response === null) {
       return null
     }
 
-    return response
+    try {
+      return JSON.parse(response)
+    } catch {
+      return response
+    }
   }
 
   async getAllFlagsAndPayloads(
@@ -308,7 +312,7 @@ class FeatureFlagsPoller {
         let matches = false
 
         if (propertyType === 'cohort') {
-          matches = matchCohort(prop, properties, this.cohorts)
+          matches = matchCohort(prop, properties, this.cohorts, this.debugMode)
         } else {
           matches = matchProperty(prop, properties)
         }
@@ -558,7 +562,8 @@ function matchProperty(
 function matchCohort(
   property: FeatureFlagCondition['properties'][number],
   propertyValues: Record<string, any>,
-  cohortProperties: FeatureFlagsPoller['cohorts']
+  cohortProperties: FeatureFlagsPoller['cohorts'],
+  debugMode: boolean = false
 ): boolean {
   const cohortId = String(property.value)
   if (!(cohortId in cohortProperties)) {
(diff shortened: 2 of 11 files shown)


posthog-node/src/feature-flags.ts
```diff
@@ -605,7 +611,9 @@ function matchPropertyGroup(
         }
       } catch (err) {
         if (err instanceof InconclusiveMatchError) {
-          console.debug(`Failed to compute property ${prop} locally: ${err}`)
+          if (debugMode) {
+            console.debug(`Failed to compute property ${prop} locally: ${err}`)
+          }
```
marandaneto, line 616: See https://github.com/PostHog/posthog-js-lite/pull/219/files#r1579057636
Could be replaced by `console.debug(`Failed to compute property ${prop} locally: ${err}`)` that internally abstracts `debugMode`

posthog-node/src/types.ts
```diff
@@ -54,7 +54,7 @@ export type PostHogFeatureFlag = {
         rollout_percentage: number
       }[]
     }
-    payloads?: Record<string, JsonType>
+    payloads?: Record<string, string>
```
marandaneto, line 57: `PostHogCoreOptions.bootstrap.featureFlagPayloads` is still `Record<string, JsonType>`, is the current code working correctly for bootstrapped flags passing `Record<string, JsonType>`? or should we update it there too?

marandaneto: Maybe `getAllFlagsAndPayloads` signature should change as well?

posthog-react-native/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Next
 
+# 3.1.1 - 2024-04-24
```
```suggestion
# 3.1.1 - 2024-04-25
```

posthog-node/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+
+# 4.0.1 - 2024-04-24
```
marandaneto, line 2:
```suggestion
# 4.0.1 - 2024-04-25
```

marandaneto:
Missing changelog for https://github.com/PostHog/posthog-js-lite/blob/main/posthog-web/CHANGELOG.md

posthog-node/src/types.ts
neilkakkar: Yep the bootstrapped ones are the parsed ones, so the type is ok and works as you'd expect.

`getAllFlagsAndPayloads` returns JsonType because this is parsed.

The type I'm updating references what we get from local eval calls, which is not a parsed payload.

posthog-node/src/feature-flags.ts
```diff
@@ -308,7 +312,7 @@ class FeatureFlagsPoller {
         let matches = false
 
         if (propertyType === 'cohort') {
-          matches = matchCohort(prop, properties, this.cohorts)
+          matches = matchCohort(prop, properties, this.cohorts, this.debugMode)
```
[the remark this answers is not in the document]
neilkakkar: oof yeah good point!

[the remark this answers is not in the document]
neilkakkar: don't want to add to this PR though 😓

neilkakkar:
thanks, added!

marandaneto (approved):

PR #236, opened 2024-06-13
fix: should attempt localstorage correctly before falling back to session storage

haddowg:
## Problem

Usage in the browser will always fallback to session storage rather than localstorage due to a simple logical conditional error.
When local storage is available it should be used when requested before falling back to sessionstorage.

## Changes

Corrects faulty condition so availability of local storage is evaluated correctly

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [x] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- use localstorage over sessionstorage where available

### DIFF
--- a/posthog-web/CHANGELOG.md
+++ b/posthog-web/CHANGELOG.md
@@ -1,4 +1,10 @@
-# Next
+# 3.0.2 - 2024-06-14
+
+## Fixed
+
+1. Fixed and error that prevented localstorage from ever being used and instead falling back to sessionstorage for persistence
+
+## Changed
 
 1. chore: change host to new address.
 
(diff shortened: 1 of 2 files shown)


marandaneto:
@benjackwhite I think this is right but just double-checking since I don't know much about the so many storage options in the browser =D

marandaneto (approved):

benjackwhite (approved):
Oh wow yeah thats a mistake. Good find!

marandaneto:
@haddowg thanks for the PR
Mind adding an entry to the changelog https://github.com/PostHog/posthog-js-lite/blob/main/posthog-web/CHANGELOG.md so we can merge and release that?

haddowg:
@marandaneto hows that?

PR #249, opened 2024-08-19
fix(feature flags) : Local evaluation : guard for null values when comparing person Properties

Phanatic:
## Problem

The current behavior of stringifying `undefined` and `null` values to compare with operators can lead to invalid results in local evaluation. 

This causes local evaluation to return `true`, while the `/decide` API does the right thing and returns false for `undefined` values in these properties.  

Customer reported that local evaluation always evaluates a condition to true, when it should evaluate to false. 
The conditions in question are : 

``` json
[
    {
        "key": "latestBuildVersion",
        "type": "person",
        "value": ".+",
        "operator": "regex"
    },
    {
        "key": "latestBuildVersionMajor",
        "type": "person",
        "value": "23",
        "operator": "gt"
    },
    {
        "key": "latestBuildVersionMinor",
        "type": "person",
        "value": "31",
        "operator": "gt"
    },
    {
        "key": "latestBuildVersionPatch",
        "type": "person",
        "value": "0",
        "operator": "gt"
    }
]
```

And the person properties passed into local evaluation are : 

``` typescript
{
    "personProperties": {
        "latestBuildVersion": undefined,
        "latestBuildVersionMajor": undefined,
        "latestBuildVersionMinor": undefined,
        "latestBuildVersionPatch": undefined
    } as unknown as Record<string, string> 
}
```

Since we stringify all of the `undefined` values, the above call returns `true` when it should return `false` because the property values are undefined and not the string `undefined`. 

## Changes

Add a `NULL_VALUES_ALLOWED_OPERATORS` array of operators that allow NULL/UNDEFINED values which is just `is_not` right now . Any other operator will fail if the overrideValue is `null` or `undefined`

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Local feature flag evaluation will return false for `undefined` or `null` property values.

### DIFF
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -7,6 +7,7 @@ import fetch from './fetch'
 // eslint-disable-next-line
 const LONG_SCALE = 0xfffffffffffffff
 
+const NULL_VALUES_ALLOWED_OPERATORS = ['is_not']
 class ClientError extends Error {
   constructor(message: string) {
     super()
@@ -305,7 +306,11 @@ class FeatureFlagsPoller {
     properties: Record<string, string>
   ): boolean {
     const rolloutPercentage = condition.rollout_percentage
-
+    const warnFunction = (msg: string) => {
+        if (this.debugMode) {
+          console.warn(msg)
+        }
+    }
     if ((condition.properties || []).length > 0) {
       for (const prop of condition.properties) {
         const propertyType = prop.type
@@ -314,7 +319,7 @@ class FeatureFlagsPoller {
         if (propertyType === 'cohort') {
           matches = matchCohort(prop, properties, this.cohorts, this.debugMode)
         } else {
-          matches = matchProperty(prop, properties)
+          matches = matchProperty(prop, properties, warnFunction)
         }
 
         if (!matches) {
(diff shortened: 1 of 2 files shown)


posthog-node/src/feature-flags.ts
```diff
@@ -474,6 +479,15 @@ function matchProperty(
   }
 
   const overrideValue = propertyValues[key]
+  if (overrideValue == null && !NULL_VALUES_ALLOWED_OPERATORS.includes(operator)) {
+    // if the value is null, just fail the feature flag comparison
+    // this isn't an InconclusiveMatchError because the property value was provided.
+    if (warnFunction) {
+      warnFunction(`Property ${key}  cannot have a value of null/undefined with the ${operator} operator`)
```
dmarticus, line 486:
```suggestion
      warnFunction(`Property ${key} cannot have a value of null/undefined with the ${operator} operator`)
```
(unless the space was intentional...)

dmarticus (approved):
looks good to me, one minor formatting change!  Thanks for the deep dive! Wonder if any of our other SDKs have this problem...

Issue #255, opened 2024-09-09
Does not work in Next.js serverless (API route)


antoinerousseau:
## Bug description

SSR support was supported in https://github.com/PostHog/posthog-js-lite/pull/57), but as mentioned [here](https://github.com/PostHog/posthog-js-lite/pull/57#discussion_r1491381624), the web version gets loaded instead of the Node.js version, and it tries to access the `window` object, throwing a `ReferenceError: window is not defined` from [here](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-web/src/posthog-web.ts#L22), [here](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-web/src/posthog-web.ts#L53) or [here](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-web/src/posthog-web.ts#L71).

## How to reproduce

1. Create a [Next.js app router API route](https://nextjs.org/docs/pages/building-your-application/routing/api-routes)
2. 
```js
import { PostHog } from "posthog-js-lite"
export async function GET() {
  const posthog = new PostHog(...)
  posthog.capture(...)
}
```
3. Call the API route from a component: you get a `ReferenceError: window is not defined`

## Related sub-libraries

- [ ] All of them
- [x] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

## Additional context

Related: https://github.com/PostHog/posthog-js-lite/issues/67#issuecomment-1488735973

benjackwhite:
This lib really isn't designed for the use case in mind. Next-js router API is running server side and as such is basically a node-js environment and so should use the built `posthog-node` target instead as that way it includes concepts such as graceful shutdown and the correct fetch implementation for your node environment.

PR #256, opened 2024-09-10
fix: check if window is available before using on web env

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js-lite/issues/255
Similar to https://github.com/PostHog/posthog-js-lite/pull/57

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [x] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X


posthog-web/src/storage.ts
```diff
@@ -93,9 +93,6 @@ const createStorageLike = (store: any): PostHogStorage => {
 }
 
 const checkStoreIsSupported = (storage: PostHogStorage, key = '__mplssupport__'): boolean => {
-  if (!window) {
-    return false
-  }
```
marandaneto, line 98: `checkStoreIsSupported` is only called if there's a `window`

posthog-web/src/context.ts
```diff
@@ -1,9 +1,9 @@
 import { utils } from '../../posthog-core'
 import { version } from '../package.json'
 
-export function getContext(window: Window): any {
+export function getContext(window: Window | undefined): any {
   let context = {}
-  if (window.navigator) {
+  if (window && window.navigator) {
```
benjackwhite, line 6: Most of these can be better written as `window?.navigator` 
```suggestion
  if (window?.navigator) {
```

posthog-web/src/context.ts
```diff
@@ -114,7 +115,10 @@ function browserVersion(userAgent: string, vendor: string, opera: boolean): numb
   return parseFloat(matches[matches.length - 2])
 }
 
-function os(window: Window): string {
+function os(window: Window | undefined): string {
+  if (!window || !window.navigator) {
```
benjackwhite, line 119:
```suggestion
  if (!window?.navigator) {
```

posthog-web/src/posthog-web.ts
```diff
@@ -19,14 +19,20 @@ export class PostHog extends PostHogCore {
...
 
     if (options?.preloadFeatureFlags !== false) {
       this.reloadFeatureFlags()
     }
   }
 
+  _getWindow(): Window | undefined {
+    const _window: Window | undefined = typeof window !== 'undefined' ? window : undefined
+    return _window
+  }
```
benjackwhite, line 34: This shouldn't be a class method but rather a util function I think

posthog-web/src/posthog-web.ts
```diff
@@ -50,6 +56,9 @@ export class PostHog extends PostHogCore {
   }
 
   fetch(url: string, options: PostHogFetchOptions): Promise<PostHogFetchResponse> {
+    // TODO: what to do here?
+    // should we move this to core? https://github.com/PostHog/posthog-js-lite/blob/main/posthog-node/src/fetch.ts
+    // and reuse it here? if window isn't available?
```
benjackwhite, line 61: I don't really get what this means. How can we fetch if there is no window?

benjackwhite (changes requested):

posthog-web/src/posthog-web.ts
marandaneto: In both Service Workers and Web Workers, there is no Window object, but the fetch() function is still available globally as part of the worker environment.
So I was assuming that https://github.com/PostHog/posthog-js-lite/blob/main/posthog-node/src/fetch.ts does something similar and we could reuse it here as a fallback if `window` isn't available.
Does that make sense?

benjackwhite: Ah i see. Yeah then I guess that [same line](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-node/src/fetch.ts#L14-L17) would I think work for loading fetch from the global instead of window? 

Might be a bit tricky to test...

posthog-web/src/posthog-web.ts
```diff
@@ -9,6 +9,11 @@ import { PostHogStorage, getStorage } from './storage'
 import { version } from '../package.json'
 import { PostHogOptions } from './types'
 
+export function _getWindow(): Window | undefined {
+  const _window: Window | undefined = typeof window !== 'undefined' ? window : undefined
+  return _window
```
pauldambra, line 14: very nit-picky but `_window` can be in-lined here
```suggestion
  return  typeof window !== 'undefined' ? window : undefined
```

posthog-web/src/posthog-web.ts
pauldambra: yep, that looks like the fix to me...

benjackwhite (approved):

marandaneto:
Screenshot 2025-01-17 at 11 01 48

works for web and node, assuming this will work for the web workers too since all I've done is to add extra checks for the window and fetch objects.
totally forgot about this issue btw, I was skimming thru the repo and noticed the open PR, well now it's done.

PR #259, opened 2024-09-12
fix: bootstrap flags do not overwrite the current values

marandaneto:
## Problem

https://posthoghelp.zendesk.com/agent/tickets/17685

## Changes

bootstrap values are always overwriting everything even if there are new values cached

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X


posthog-react-native/CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # Next
 
+# 3.2.0 - 2024-09-12
+
 ## Changed
 
 1. chore: default `captureMode` changed to `json`.
     1. To keep using the `form` mode, just set the `captureMode` option to `form` when initializing the PostHog client.
+2. fix: fix: bootstrap flags do not overwrite the current values
```
marandaneto, line 9:
```suggestion
2. fix: bootstrap flags do not overwrite the current values
```

posthog-core/src/index.ts
```diff
@@ -724,25 +724,54 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
-    if (options?.bootstrap?.distinctId) {
-      if (options?.bootstrap?.isIdentifiedId) {
-        this.setPersistedProperty(PostHogPersistedProperty.DistinctId, options.bootstrap.distinctId)
+    const bootstrap = options?.bootstrap
+    if (!bootstrap) {
+      return
+    }
+
+    // bootstrap options are only set if no persisted values are found
+    // this is to prevent overwriting existing values
```
marandaneto, line 733: this code lives in the core package so I assume this affects node and web as well.
not sure if node and web will have storage (likely memory storage), but IMO the change still makes sense.
if you bootstrap a flag and either there's a value or a new value comes from the decide API, the cached value or the new value from flags have precedence over the bootstrap flag.

pauldambra: hmmm, I assume that's checked with @PostHog/team-feature-success my assumption would have been that client provided bootstrap values would override remote values - like we do with config. E.g. disabled with client config overrules enabled with remote config.

but that's only asked with no context - I don't know the right answer

pauldambra (approved):
logic looks correct for the goal... just the open "product" question for me. once we know the canonical answer we should update https://posthog.com/docs/feature-flags/bootstrapping (which assumes i'm not the only person that would make the same assumption 🙈)

marandaneto:
> logic looks correct for the goal... just the open "product" question for me. once we know the canonical answer we should update [posthog.com/docs/feature-flags/bootstrapping](https://posthog.com/docs/feature-flags/bootstrapping) (which assumes i'm not the only person that would make the same assumption 🙈)

I think its correct already:

> To have your feature flags available immediately, you can initialize PostHog with precomputed values until it has had a chance to fetch them. This is called bootstrapping.

that means, after having the chance to fetch them, use the fetched values and not the bootstraped values anymore, which is what this fix does.

posthog-core/src/index.ts
marandaneto: https://github.com/PostHog/posthog-js-lite/pull/260/files that would work for RN only (would need fix tests) but if that's the expectation, happy to do it, I still think it's strange as per [docs](https://github.com/PostHog/posthog-js-lite/pull/259#issuecomment-2346441186), and as I have used bootstrap flags in other tools.
@Phanatic @neilkakkar can you decide on the behaviour here? thx

Phanatic (approved):
This looks good, but instead of checking every value before setting it here : 
https://github.com/PostHog/posthog-js-lite/pull/259/files#diff-e8ba078a1fb3324f02789b3d54af7bd54e24168943fcb1ffa6de45486fa6d760R742

Would it make sense to change the behavior of `setPersistedProperty` to change to be `trySetPersistedProperty` which will do nothing if the value is already set? 

or is this logic special to only `distinctId` and `anonymousId` ?

marandaneto:
> #259 (files)

Special only to these 2, those are unique identifiers per installation or until the user `identify` or `reset`.
eg bootstrap with `distinctId` A, then the user calls `identify` with `distinctId` B, the app restarts, and the `distinctId` should still be `B` and not the bootstrapped `A` anymore because the user already identified.

Issue #268, opened 2024-09-21
RN Session replay plugin isn't starting a new recording after app restart


marandaneto:
## Bug description

Context https://github.com/PostHog/posthog/issues/13269#issuecomment-2363950067
This is by design but I understand why it isn't obvious.
On RN, the session only rotates if the app is inactive for more than 30 minutes, and the session ID survives app restart which is different than Android and iOS.
Changing this would break the behavior of people who rely on this for session durations.
So we have a few options here, change the behavior and break compatibility, add a flag to opt in to the new behavior, or else.

## How to reproduce

1.
2.
3.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

marandaneto:
cc @chasem-dev
https://github.com/PostHog/posthog-js-lite/releases/tag/posthog-react-native-v3.3.0

chasem-dev:
I found where it looks like you exposed the variable:
https://github.com/PostHog/posthog-ios/commit/501cc8e48991747391240475e9c8cc341e995b08

It seems like the version bump maybe wasn't picked up in a release?
https://github.com/PostHog/posthog-ios/commit/333ca7fa6d5072b1fa12f93226c2e8ea153f3b45

marandaneto:
@chasem-dev have you upgraded your pods? Try to clean and install your pods, the session replay plugin uses the iOS SDK

chasem-dev:
@marandaneto That worked! I ran `cd ios` and then did pod update. Testing things out now. Sorry for the false alarm! Thanks for the help

PR #271, opened 2024-09-24
changelog

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js-lite/issues/268

### Bump level

- [ ] Major
- [X] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/src/posthog-rn.ts
+++ b/posthog-react-native/src/posthog-rn.ts
@@ -57,6 +57,14 @@ export type PostHogOptions = PostHogCoreOptions & {
    * Experimental support
    */
   sessionReplayConfig?: PostHogSessionReplayConfig
+
+  /**
+   * The session id ($session_id) will be persisted across app restarts.
+   * This is an option for back compatibility (if enabled), so your current data isn't skewed with the new version of the SDK.
+   * If this is false, the session id will be reset on app restart.
+   * Defaults to false
+   */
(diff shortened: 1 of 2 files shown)


posthog-react-native/src/posthog-rn.ts
```diff
@@ -57,6 +57,14 @@ export type PostHogOptions = PostHogCoreOptions & {
...
    */
   sessionReplayConfig?: PostHogSessionReplayConfig
+
+  /**
+   * If enabled, the session id ($session_id) will be persisted across app restarts.
+   * This is an option for back compatibility, so your current data isn't skewed with the new version of the SDK.
+   * If this is false, the session id will be always reset on app restart.
+   * Defaults to false
+   */
+  enablePersistSessionIdAcrossRestart?: boolean
```
marandaneto, line 67: Another option is `enableRotateSessionIdAcrossRestart` which is enabled by default, and open to ideas.

pauldambra: It should be a safe enough change that opting people into the new (and more desirable behaviour) is acceptable

So I like the people have the option to opt back into the old behaviour but that we force upgrade to the new behaviour 👍

pauldambra (approved):
👍

PR #278, opened 2024-10-07
Remove Debug Statement when SessionReplay doesn't load

gja:
This brings the structure in line with the other optional dependencies, ensuring that logs stay quiet

## Problem

When SessionReplay isn't installed, it produces a warning log. This is inconsistent with the other optional dependencies.

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Removed a warning log when posthog-react-native-session-replay is not installed


marandaneto:
Makes sense @gja 
Thanks

marandaneto (approved):

PR #284, opened 2024-10-14
fix: only logging if debug enabled

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-react-native-session-replay/issues/7

## Changes

Only logs message if debug is enabled

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [x] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -113,11 +113,15 @@ export abstract class PostHogCoreStateless {
     this._isInitialized = true
   }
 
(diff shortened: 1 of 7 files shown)


pauldambra (approved):
https://github.com/PostHog/posthog-js/blob/main/src/utils/logger.ts

could do something like that but not blocking :)

PR #291, opened 2024-10-19
fix: all sdkReplayConfig should have a default value

marandaneto:
## Problem

https://posthog.com/questions/installing-session-replay-and-updating-posthog-crashes-the-app

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/CHANGELOG.md
+++ b/posthog-react-native/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Next
 
+# 3.3.6 - 2024-10-19
+
+1. fix: all sdkReplayConfig should have a default value
+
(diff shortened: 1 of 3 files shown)


marandaneto:
> Not fluent in js but I think we could also use the spread operator to merge defaults with params.
> 
> LGTM though

Yes, the spread operator would merge but it'd have the wrong result if someone passes eg `maskAllTextInputs: undefined`, the undefined would have precedence over the default value, with the current form, no, the default value would have precedence.

PR #294, opened 2024-10-24
fix: session replay respects the disabled flag

marandaneto:
## Problem

Addresses part of https://github.com/PostHog/posthog-react-native-session-replay/issues/10

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/tsconfig.json
+++ b/posthog-react-native/tsconfig.json
@@ -5,7 +5,7 @@
     "esModuleInterop": true,
     "jsx": "react-native",
     "lib": ["ESNext"],
-    "module": "ESNext",
+    "module": "node16",
     "moduleResolution": "node16",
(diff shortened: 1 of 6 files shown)


posthog-react-native/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "esModuleInterop": true,
     "jsx": "react-native",
     "lib": ["ESNext"],
-    "module": "ESNext",
```
marandaneto, line 8: VS warned that if moduleResolution is node16, module should be too https://www.typescriptlang.org/tsconfig/#moduleResolution

PR #305, opened 2024-11-12
fix: Enforce the shutdown timeout 

benjackwhite:
## Problem

The shutdown timeout is misleading. If a fetch request is timing out we may still sit on it as we only do the check after it is done.

## Changes

Change to a timeout based solution. It's a little complex but i think thats just _the way_.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Shutdown will now respect the timeout and forcefully return rather than returning after the next fetch.

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -708,14 +708,14 @@ export abstract class PostHogCoreStateless {
     }
 
     return Promise.race([
-      doShutdown(),
       new Promise<void>((_, reject) => {
         safeSetTimeout(() => {
           this.logMsgIfDebug(() => console.error('Timed out while shutting down PostHog'))
           hasTimedOut = true
-          reject()
+          reject('Timeout while shutting down PostHog. Some events may not have been sent.')
         }, shutdownTimeoutMs)
       }),
+      doShutdown(),
     ])
   }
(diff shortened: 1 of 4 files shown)


marandaneto (approved):

marandaneto:
@benjackwhite if you add changelog - update version we can cut a version o/

posthog-core/src/index.ts
```diff
@@ -672,39 +672,51 @@ export abstract class PostHogCoreStateless {
...
+        this.logMsgIfDebug(() => console.error('Error while shutting down PostHog', e))
       }
-    } catch (e) {
-      if (!isPostHogFetchError(e)) {
-        throw e
-      }
-      this.logMsgIfDebug(() => console.error('Error while shutting down PostHog', e))
     }
+
+    return Promise.race([
```
marandaneto, line 710: > Shutdown will loop over and flush data but outside of issues with our ingestion endpoint I can't see why it would hang

Yeah same, I did not expect this happening but now it will bail out regardless.

PR #306, opened 2024-11-13
fix: respect the given propsToCapture autocapture option

marandaneto:
## Problem

follow up https://github.com/PostHog/posthog-js-lite/pull/302/files

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/src/autocapture.tsx
+++ b/posthog-react-native/src/autocapture.tsx
@@ -47,8 +47,8 @@ export const autocaptureFromTouchEvent = (e: any, posthog: PostHog, options: Pos
     customLabelProp = defaultPostHogLabelProp,
     maxElementsCaptured = 20,
     ignoreLabels = [],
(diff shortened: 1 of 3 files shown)


posthog-react-native/src/autocapture.tsx
```diff
@@ -47,8 +47,8 @@ export const autocaptureFromTouchEvent = (e: any, posthog: PostHog, options: Pos
     customLabelProp = defaultPostHogLabelProp,
     maxElementsCaptured = 20,
     ignoreLabels = [],
+    propsToCapture = ['style', 'testID', 'accessibilityLabel', customLabelProp, 'children'],
```
marandaneto, line 50: propsToCapture from options have precedence over the default values

PR #307, opened 2024-11-15
fix: session replay forces the session id if the SDK is already enabled

marandaneto:
## Problem

for some reason, sometimes but not so rarely, the recording does not have associated events but the SDK did capture events, and the session id does not match.
I believe the SDK can be already enabled or the session id rotated wrongly in the native SDK.
Now the hybrid SDK fully controls the session id lifecycle and if the SDK is somehow already enabled, we force the session id again.
Depends on https://github.com/PostHog/posthog-ios/pull/253 as well

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/examples/example-expo-latest/package.json
+++ b/examples/example-expo-latest/package.json
@@ -21,7 +21,7 @@
     "expo-localization": "~15.0.3",
     "expo-status-bar": "~1.12.1",
     "posthog-react-native": "file:.yalc/posthog-react-native",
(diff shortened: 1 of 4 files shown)


pauldambra (approved):
low context 👍

PR #308, opened 2024-11-17
fix: update Sentry Integration

tsekiguchi:
## Problem

In the current version of Sentry, they have changed their Integration type. In the current state, when trying to add the Posthog Integration to Sentry it's throwing an error:

```
Type 'PostHogSentryIntegration' is not assignable to type 'Integration'.
  Types of property 'setupOnce' are incompatible.
    Type '(addGlobalEventProcessor: (callback: any) => void, getCurrentHub: () => any) => void' is not assignable to type '() => void'.
      Target signature provides too few arguments. Expected 2 or more, but got 0.ts(2322)
```

[A couple other users on the PostHog integration page have also been running into this issue](https://posthog.com/docs/libraries/node) mentioned by Ariel and Vitor.

This is related to two breaking changes in Sentry's updated Integration:
1. The `setupOnce()` method no longer passes any functions.
2. `getCurrentHub()` is being [deprecated](https://github.com/getsentry/sentry-javascript/blob/develop/MIGRATION.md#deprecate-getcurrenthub)

The PostHog sentry-integration now needs a different way to get the projectId and add events coming from Sentry, which are used in capture events in posthog.

Last thing I'll say is that this is my first open source contribution! So apologies if something is messy or not right :) Please let me know if there's anything I can improve!

## Changes

Since the method `setupOnce()` no longer passes the current functions used for events and hub, we need to use `setup()` which passes the Sentry `client()` class. Using that class, we have access to `addEventProcessor(event)` function and `getDsn().projectId` that was previously being used. That satisfies the information needed by PostHog.

We also no longer need the specific _SentryIntegration interface.

I've also updated the tests to account for the new `setup()` class with the `addEventProcessor()` function. It's currently passing on my local machine.

## Release info Sub-libraries affected

As noted in the integration, we don't import `@sentry/types`, but the latest ones were used to check everything in development.

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Fixed the PostHog Sentry Integration to support the latest version of Sentry (v8)

### DIFF
--- a/posthog-node/package.json
+++ b/posthog-node/package.json
@@ -27,6 +27,7 @@
     "rusha": "^0.8.14"
   },
   "devDependencies": {
+    "@sentry/types": "^8.38.0",
     "@types/node": "^18.0.0",
     "commander": "^9.3.0"
   },
--- a/posthog-node/src/extensions/sentry-integration.ts
+++ b/posthog-node/src/extensions/sentry-integration.ts
@@ -8,25 +8,19 @@ import { type PostHog } from '../posthog-node'
 
 // import {
 //     Event as _SentryEvent,
-//     EventProcessor as _SentryEventProcessor,
 //     Exception as _SentryException,
-//     Hub as _SentryHub,
 //     Integration as _SentryIntegration,
 //     Primitive as _SentryPrimitive,
+//     Client as _SentryClient,
 // } from '@sentry/types'
 
 // Uncomment the above and comment the below to get type checking for development
 
 type _SentryEvent = any
-type _SentryEventProcessor = any
-type _SentryHub = any
 type _SentryException = any
 type _SentryPrimitive = any
-
-interface _SentryIntegration {
-  name: string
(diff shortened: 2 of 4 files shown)


marandaneto:
@tsekiguchi there was already a [similar PR](https://github.com/PostHog/posthog-js-lite/pull/243)
something for @PostHog/team-error-tracking

tsekiguchi:
Ah I didn't see that! And I'm seeing that one has a way to differentiate for Sentry v7. I see that he submitted that PR about a month ago, is there any update on when it might be merged?

daibhin:
Hi @tsekiguchi, I managed to work on this today: https://github.com/PostHog/posthog-js-lite/pull/311. Just waiting on a review and then hopefully you will be able to use Sentry v8. We're also working on an error tracking product right now. Feel free to email me if you're interested in early access to that :)

PR #311, opened 2024-11-20
chore: support Sentry v8

daibhin:
## Problem

Our Sentry integration does not work for v8. Prompted by the great work in https://github.com/PostHog/posthog-js-lite/pull/243.

## Changes

We already support v8 in `posthog-js` so I went about copying the approach used there and fixing any of the drift between the two implementations of the Sentry integration. Because of the way that the `posthog-node` package works the code isn't 100% equivalent but this works pretty well

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for Sentry v8 integration

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,6169 @@
+lockfileVersion: '6.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+devDependencies:
+  '@babel/core':
+    specifier: ^7.18.6
+    version: 7.26.0
+  '@babel/preset-env':
+    specifier: ^7.18.6
+    version: 7.26.0(@babel/core@7.26.0)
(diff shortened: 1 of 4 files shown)


oliverb123 (approved):

RickVM:
Took me some digging in the commits to find how to get v8 working as I didn't see it in the docs either. Leaving it here for others reference:

```ts
import { sentryIntegration } from 'posthog-node';

// Initialize posthog, and import it here
import posthog from '@/lib/posthog/server';

Sentry.init({
  dsn: 'sentry_dsn',
  integrations: [
    sentryIntegration(posthog, {
      organization: 'your-org,
      projectId: 12345
    })
  ]
});

```

Issue #312, opened 2024-11-21
(React Native) posthog.identify types do not match docs


CallumHemsley:
## Bug description
[Docs](https://posthog.com/docs/libraries/react-native#identifying-users) contain the following example:
```js
posthog.identify('distinctID', 
  { // userProperties:
      email: 'user@posthog.com',
      name: 'My Name'
  },
  { // userPropertiesSetOnce:
      date_of_first_log_in: '2024-03-01'
  }
)
```
However, I can see the posthog types from `posthog-react-native` for identity is: ```identify(distinctId?: string, properties?: PostHogEventProperties, options?: PostHogCaptureOptions): void;```

I guess aside from updating the docs, how can we achieve the userPropertiesSetOnce in RN, or is it just the types that are broken? 

## How to reproduce

1. install latest version following docs e.g `npx expo install posthog-react-native expo-file-system expo-application expo-device expo-localization`
2. See in package json you have `"posthog-react-native": "^3.3.13",`
3. Call `usePostHog` with following code below, and right click into types for identify

```js
import { usePostHog } from "posthog-react-native";
import {
  PostHogCaptureOptions,
  PostHogEventProperties,
} from "posthog-react-native/lib/posthog-core/src";

export const useAnalytics = () => {
  const posthog = usePostHog();

  const capture = (
    event: string,
    properties?: {
      [key: string]: unknown;
    },
    options?: PostHogCaptureOptions,
  ) => {
    posthog?.capture(event, properties);
  };

  const identify = (
    distinctId?: string,
    properties?: PostHogEventProperties,
    options?: PostHogCaptureOptions,
  ) => {
    // posthog?.identify(distinctId, properties, options);
    posthog.identify(
      "distinctID",
      {
        // userProperties:
        email: "user@posthog.com",
        name: "My Name",
      },
      {
        // userPropertiesSetOnce:
        date_of_first_log_in: "2024-03-01",
      },
    );
  };

  return { capture, identify };
};
```

## Related sub-libraries

- [X] posthog-react-native

marandaneto:
Hello @CallumHemsley thanks for this issue
The `identify` method does not have support for the `userPropertiesSetOnce` ($set_once) option, I'm gonna fix that.
https://posthog.com/docs/libraries/react-native#set_once

marandaneto:
https://github.com/PostHog/posthog-js-lite/pull/313

PR #313, opened 2024-11-21
chore: identify set once

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js-lite/issues/312

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [X] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Added support for X


posthog-core/test/posthog.identify.spec.ts
```diff
@@ -30,7 +30,6 @@ describe('PostHog Core', () => {
             properties: {
               $lib: 'posthog-core-tests',
               $lib_version: '2.0.0-alpha',
-              foo: 'bar',
```
marandaneto, line 33: this is under `$set` and not dupe as a high level prop as well

posthog-core/src/index.ts
```diff
@@ -953,10 +953,17 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         this.groups(properties.$groups)
       }
 
+      // promote $set and $set_once to top level
+      const userPropsOnce = properties?.$set_once
+      delete properties?.$set_once
+
+      // if no $set is provided we assume all properties are $set
+      const userProps = properties?.$set || properties
```
marandaneto, line 961: Hacky workaround to keep back-compatibility, ideally those 2 Dicts were additional method parameters instead of a single parameter typed as a generic key/value that can be anything.

benjackwhite (approved):

CallumHemsley:
Just wanted to comment to say thanks for jumping on this so fast! Have tested and it's working great 🎉

PR #315, opened 2024-11-21
fix(flags): send `feature_flag_called` events with correct payload data in the event that we need to fetch the payloads from the server

dmarticus:
Follow on to [this discussion](https://github.com/PostHog/posthog-python/pull/143#discussion_r1851492487) and [this slack thread](https://posthog.slack.com/archives/C07Q2U4BH4L/p1732202609363039?thread_ts=1732144656.105669&cid=C07Q2U4BH4L), it looks like the bug with the Python SDK (where we send empty responses in our `feature_flag_called` events when we call `get_feature_flag_payload`) is also present in the Node SDK.  

This change makes it so that we ensure that we have a relevant response + payload to send on the event body whenever we fall back to remotely evaluating the feature flag payload.

### DIFF
--- a/posthog-node/src/posthog-node.ts
+++ b/posthog-node/src/posthog-node.ts
@@ -287,14 +287,7 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
       disableGeoip?: boolean
     }
   ): Promise<JsonType | undefined> {
-    const {
-      groups,
-      disableGeoip,
-      onlyEvaluateLocally = false,
-      sendFeatureFlagEvents = true,
-      personProperties,
-      groupProperties,
-    } = options || {}
+    const { groups, disableGeoip, onlyEvaluateLocally = false, personProperties, groupProperties } = options || {}
 
     const { allPersonProperties, allGroupProperties } = this.addLocalPersonAndGroupProperties(
       distinctId,
--- a/posthog-node/src/posthog-node.ts
+++ b/posthog-node/src/posthog-node.ts
@@ -342,7 +342,6 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
       fetchedOrLocalPayloads = fetchedData.payloads || {}
     }
 
-    // Extract the final response and payload for the specific key
     const finalResponse = fetchedOrLocalFlags[key]
     const finalPayload = fetchedOrLocalPayloads[key]
     const finalLocallyEvaluated = payloadWasLocallyEvaluated
@@ -361,7 +360,6 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
       disableGeoip,
     })
 
-    // Return the final payload
     return finalPayload
   }
 
--- a/posthog-node/src/posthog-node.ts
+++ b/posthog-node/src/posthog-node.ts
@@ -303,58 +303,72 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
       groupProperties
     )
 
-    const effectiveMatchValue =
-      matchValue !== undefined
-        ? matchValue
-        : await this.getFeatureFlag(key, distinctId, {
-            ...options,
-            onlyEvaluateLocally: true,
-            sendFeatureFlagEvents: false,
-          })
-
-    const { response, payload } = effectiveMatchValue
-      ? {
-          response: effectiveMatchValue,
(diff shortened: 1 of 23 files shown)


posthog-node/src/posthog-node.ts
```diff
@@ -287,59 +287,82 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
...
-
-    let response = undefined
-
-    // Try to get match value locally if not provided
-    if (!matchValue) {
+    if (matchValue === undefined) {
       matchValue = await this.getFeatureFlag(key, distinctId, {
         ...options,
         onlyEvaluateLocally: true,
+        sendFeatureFlagEvents: false,
```
dmarticus, line 303: explicitly don't send events when we evaluate the feature flag (since these events won't have response data for the payload), we'll send them manually from this method insstead.

posthog-node/test/posthog-node.spec.ts
```diff
@@ -919,6 +924,34 @@ describe('PostHog Node.js', () => {
...
+          properties: expect.objectContaining({
+            $feature_flag: 'beta-feature',
+            $feature_flag_response: true,
+            $feature_flag_payload: { variant: 'A' },
+            locally_evaluated: true,
+            [`$feature/${'beta-feature'}`]: true,
+          }),
+        })
+      )
+      mockedFetch.mockClear()
```
dmarticus: added a test case that the `getFeatureFlagPayload` method also sends a `feature_flag_called` event.

havenbarnes (approved):
Changes make sense!

posthog-node/src/posthog-node.ts
```diff
@@ -287,59 +287,80 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
...
+    const finalPayload = fetchedOrLocalPayloads[key]
+    const finalLocallyEvaluated = payloadWasLocallyEvaluated
+
+    this.capture({
+      distinctId,
+      event: '$feature_flag_called',
+      properties: {
+        $feature_flag: key,
+        $feature_flag_response: finalResponse,
+        $feature_flag_payload: finalPayload,
```
marandaneto, line 355: is this flag only for local evaluation? I don't think we send this one for any other Mobile SDK, but we do support feature flag payloads

marandaneto:
@dmarticus for [releasing](https://github.com/PostHog/posthog-js-lite/blob/main/RELEASING.md).

posthog-node/src/posthog-node.ts
marandaneto: What I meant is that I did not see this new property `feature_flag_payload` anywhere else yet, is that new? if so, we'd need to add to other SDKs as well?
JS for example has `$feature_flag_payloads` but its not used when capturing a `$feature_flag_called` event

marandaneto: So my question about local evaluation is if this is only for backend SDKs (only backend SDKs have local evaluation afaik), because I have not found any reference on the JS SDK

dmarticus: AH I see!  Okay this is a good point to raise: yes; right now this fallback on happens for local evaluation on backend SDKs, and we don't include any payload information in these events.  In fact, I don't even think we send `feature_flag_called` events whenever a client-side SDK is used to get payloads, the code just goes like this
```
    getFeatureFlagPayload(key: string): JsonType {
        const payloads = this.getFlagPayloads()
        return payloads[key]
    }
```
where `getFlagPayloads` just loads the payloads from persistence.
```
    getFlagPayloads(): Record<string, JsonType> {
        const flagPayloads = this.instance.get_property(PERSISTENCE_FEATURE_FLAG_PAYLOADS)
        return flagPayloads || {}
    }
```

So this is already feature gap between client-side and server side SDKs in more ways than just including the payload – our server-side SDKs send `feature_flag_called` events on both `getFeatureFlag` and `getFeatureFlagPayload` calls.  I feel like I should make this documentation clear somewhere if it isn't already, since users could get confused.

dmarticus: yeah, I think that's a common pattern for the client-side SDKs that the server-side ones may not follow as much (e.g., the reason I started working on this change at all was because we have a user that's using our python SDK only for the `get_feature_flag_payload` method).

Okay, so here's what I'm going to do going forward:

- ship this change to the node SDK
- keep posthog-js as is

Thanks for chatting about it with me!

Issue #314, opened 2024-11-21
Data discrepancy between bootstrapped data and front-end data


bjfresh:
## Bug description

Data discrepancy between bootstrapped data and front-end data

## How to reproduce

On server:
```ts
import { z } from "zod"
import { rpc } from "../../trpc"
import { PostHogClient } from "../../../clients/posthog"

import { PostHog } from "posthog-node"
import { POSTHOG_API_KEY } from "../constants"

export const PostHogClient = new PostHog(POSTHOG_API_KEY, {
	host: "https://us.i.posthog.com",
	flushAt: 1,
	flushInterval: 0,
})

const userFlagsSchema = z.object({
	profileTBA: z.string(),
})

export const userFlags = rpc.input(userFlagsSchema).query(async ({ input }) => {
	const userFlags = await PostHogClient.getAllFlags(input.profileTBA)
	console.log({ userFlags })
	await PostHogClient.shutdown()

	return userFlags
})

```

Bootstrap + init React Native client:
```ts
import { trpc } from "@/services/trpc"
import { usePermaUser } from "./usePermaUser"
import { useEffect, useState } from "react"
import { initializePostHogClient } from "@/services/posthog"
import type PostHog from "posthog-react-native"

export function usePostHogClient({
	profileTBA,
}: {
	profileTBA?: `0x${string}`
}): PostHog | null {
	const [PostHogClient, setPosthogClient] = useState<PostHog | null>(null)
	const permaUser = usePermaUser({ userAddress: profileTBA })

	// Fetch the user's feature flags from the server so we can bootstrap the PostHog client
	// This helps ensure that the PostHog client is ready to go with accurate flag data when we need it
	// Without this, the data may not be ready in time for the first feature flag check and the flag may be inaccurate
	const userFeatureFlags = trpc.featureFlags.userFlags.useQuery(
		{
			// biome-ignore lint/style/noNonNullAssertion: <We know profileTBA is defined because this hook is otherwise disabled>
			profileTBA: profileTBA!,
		},
		{
			enabled: !!profileTBA,
		},
	)

	// console.log({ userFeatureFlags })

	// Init the PostHog client and bootstrap with the user's feature flags
	useEffect(() => {
		const init = async () => {
			if (!!PostHogClient || !userFeatureFlags.data || !profileTBA) {
				// Exit early if the PostHog client is already initialized, or if we're missing data
				return
			}

			const posthog = await initializePostHogClient({
				profileTBA,
[...]

PR #320, opened 2024-11-28
chore(flags): Add new debugging property `$feature_flag_bootstrapped_response` and `$feature_flag_bootstrapped_payload` to `$feature_flag_called` event

havenbarnes:
## Problem

When SDK users make use of feature flag bootstrapping, the response shown in the $feature_flag_called event can show different values than what would have been fetched remotely. This can lead to confusion when reviewing flag value breakdown results, for example making it seem as if a flag that is turned completely to `true` was evaluated to `false`. These two new event properties are mostly useful for debugging customer issues, but could also be used in product to show users that the response shown came from bootstrapped values, and not /decide

## Changes

Nothing functional, just additional event properties

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [x] posthog-web
- [ ] posthog-node (no server-side bootstrapping)
- [x] posthog-react-native

### Changelog notes

- Add new debugging property `$feature_flag_bootstrapped_response` and `$feature_flag_bootstrapped_payload` to `$feature_flag_called` event

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -777,6 +777,8 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         )
 
       if (Object.keys(bootstrapFlags).length) {
+        this.setPersistedProperty(PostHogPersistedProperty.BootstrapFeatureFlags, bootstrapFlags)
+
         const currentFlags =
           this.getPersistedProperty<PostHogDecideResponse['featureFlags']>(PostHogPersistedProperty.FeatureFlags) || {}
         const newFeatureFlags = { ...bootstrapFlags, ...currentFlags }
@@ -785,6 +787,8 @@ export abstract class PostHogCore extends PostHogCoreStateless {
 
       const bootstrapFlagPayloads = bootstrap.featureFlagPayloads
       if (bootstrapFlagPayloads && Object.keys(bootstrapFlagPayloads).length) {
+        this.setPersistedProperty(PostHogPersistedProperty.BootstrapFeatureFlagPayloads, bootstrapFlagPayloads)
+
         const currentFlagPayloads =
           this.getPersistedProperty<PostHogDecideResponse['featureFlagPayloads']>(
             PostHogPersistedProperty.FeatureFlagPayloads
@@ -1252,6 +1256,12 @@ export abstract class PostHogCore extends PostHogCoreStateless {
       this.capture('$feature_flag_called', {
         $feature_flag: key,
         $feature_flag_response: response,
+        $feature_flag_bootstrapped_response: this.getPersistedProperty<PostHogDecideResponse['featureFlags']>(
+          PostHogPersistedProperty.BootstrapFeatureFlags
+        )?.[key],
+        $feature_flag_bootstrapped_payload: this.getPersistedProperty<PostHogDecideResponse['featureFlagPayloads']>(
+          PostHogPersistedProperty.BootstrapFeatureFlagPayloads
+        )?.[key],
       })
     }
 
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -1264,7 +1264,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         $feature_flag_bootstrapped_payload: this.getPersistedProperty<PostHogDecideResponse['featureFlagPayloads']>(
           PostHogPersistedProperty.BootstrapFeatureFlagPayloads
         )?.[key],
-        // If we haven't gotten a response from the /decide endpoint, we must have used the bootstrapped value
+        // If we haven't yet received a response from the /decide endpoint, we must have used the bootstrapped value
         $used_bootstrap_value: !this.getPersistedProperty(PostHogPersistedProperty.DecideEndpointWasHit),
       })
     }
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -1194,6 +1194,8 @@ export abstract class PostHogCore extends PostHogCoreStateless {
                 Object.entries(newFeatureFlagPayloads || {}).map(([k, v]) => [k, this._parsePayload(v)])
               )
             )
+            // Mark that we hit the /decide endpoint so we can capture this in the $feature_flag_called event
+            this.setPersistedProperty(PostHogPersistedProperty.DecideEndpointWasHit, true)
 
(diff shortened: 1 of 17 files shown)


havenbarnes:
@dmarticus I'm thinking it may be smart to add an additional property here as well, indicated whether flags have been fetched yet or not at the time of the flag check. Something like `$feature_flags_fetched: boolean`. WDYT?

dmarticus:
> @dmarticus I'm thinking it may be smart to add an additional property here as well, indicated whether flags have been fetched yet or not at the time of the flag check. Something like `$feature_flags_fetched: boolean`. Otherwise, we don't have a clear way to know if `$feature_flag_response` matches `$feature_flag_bootstrapped_response` because bootstrapping was done correctly, or if the flags have just been fetched already. WDYT?

I love this idea; seems like it couldn't hurt to add this type of metadata to our flags requests – maybe something like `used_boostrap_value` is more clear than `feature_flags_fetched`, if the goal is to track that bootstrapping was done correctly.  And I think that could be done in a subsequent PR.  But I like this!

dmarticus (approved):
stamped with a comment!

posthog-react-native/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Next
 
+# 3.4.1 - 2024-11-27
```
marandaneto, line 3: Should this be a minor since it adds a new feature?

marandaneto:
@havenbarnes would be cool to add a test case for bootstrap flags and payloads, check the file `posthog.featureflags.spec.ts`

posthog-react-native/CHANGELOG.md
havenbarnes: @marandaneto  It's technically just for debugging ability, no plans to use in product right now. You know better here, so let me know what you think!

dmarticus: eh I think @marandaneto has a good point since it's kinda going to be used in our product (once those events get propagated to the payload, all users will have access to it).  Useful for debugging on our end, sure, but could also be handy for customers I imagine.

marandaneto: your call here, just food for though, the patch is just for bug fixes usually (semver).

havenbarnes:
Thanks @marandaneto , added some tests. Also added the `$used_bootstrap_value` property discussed above. I'm imitating same behavior as `this.instance.decideEndpointWasHit` from `posthog-js` (https://github.com/PostHog/posthog-js/blob/bootstrapped-flag-evt-property/src/posthog-featureflags.ts#L295). I'll wait for your approval!

posthog-react-native/CHANGELOG.md
havenbarnes: Makes sense for traditional semver, updated to minor bump!

posthog-react-native/CHANGELOG.md
```diff
@@ -1,13 +1,17 @@
...
+
 # 3.5.0 - 2024-12-03
 
 1. fix: deprecate maskPhotoLibraryImages due to unintended masking issues
 
 # 3.4.0 - 2024-11-26
 
 1. feat: automatically mask out user photos and sandboxed views like photo picker (iOS Only)
-  1. To disable masking set `maskAllSandboxedViews` and `maskPhotoLibraryImages` to false
+1. To disable masking set `maskAllSandboxedViews` and `maskPhotoLibraryImages` to false
```
marandaneto, line 14: I believe those changes (the rest of the file) are unintentional since its a nested bullet list, please revert

posthog-web/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Next
 
+# 3.1.1 - 2024-11-27
+
+## Changed
+
+1. Add new debugging property `$feature_flag_bootstrapped_response` and `$feature_flag_bootstrapped_payload` to `$feature_flag_called` event
```
marandaneto, line 7: we can also mention `$used_bootstrap_value`

posthog-web/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Next
 
+# 3.1.1 - 2024-11-27
```
marandaneto, line 3: this should be a minor as well similarly to RN

posthog-core/src/index.ts
```diff
@@ -1252,6 +1258,14 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
         $feature_flag: key,
         $feature_flag_response: response,
+        $feature_flag_bootstrapped_response: this.getPersistedProperty<PostHogDecideResponse['featureFlags']>(
+          PostHogPersistedProperty.BootstrapFeatureFlags
+        )?.[key],
+        $feature_flag_bootstrapped_payload: this.getPersistedProperty<PostHogDecideResponse['featureFlagPayloads']>(
+          PostHogPersistedProperty.BootstrapFeatureFlagPayloads
+        )?.[key],
+        // If we haven't yet received a response from the /decide endpoint, we must have used the bootstrapped value
+        $used_bootstrap_value: !this.getPersistedProperty(PostHogPersistedProperty.DecideEndpointWasHit),
```
marandaneto, line 1268: I think it's ok having this but you could also just compare the flag (the cached flag - likely not there if decide isn't called, and the value from the bootstrap flags), so you know if it's being used from the bootstrap or not, right?
Since using storage is not super fast, the less we need the better, your call here, but if possible I'd evaluate this at runtime.

posthog-react-native/CHANGELOG.md
havenbarnes: Weird my VS code was doing this, reverted

posthog-core/src/index.ts
havenbarnes: Are you referring to `PostHogPersistedProperty.FeatureFlags` or a platform specific cache?

I was hoping we could avoid this as well, but we store bootstrapped flags in `PostHogPersistedProperty.FeatureFlags` upon initialization, so it would already be there

marandaneto (approved):

havenbarnes:
> @havenbarnes I think you can merge/release this, that would help debugging #314 o/

Thanks @marandaneto , i was hoping to wait for to get unblocked https://github.com/PostHog/posthog-js/pull/1567 but obviously this will be helpful regardless. If I can't get it unblocked today I'll just merge this

PR #329, opened 2024-12-14
Switch from rusha to native (node:crypto) sha1 implementation

kravetsone:
## Problem

Useless [rusha](https://www.npmjs.com/package/rusha) usage

## Changes

I switch from [rusha](https://www.npmjs.com/package/rusha) to native (node:crypto) sha1 implementation

#### Node.js

```sh
$ node benchmarks/rusha-vs-native.mjs
clk: ~3.99 GHz
cpu: AMD Ryzen 7 7700 8-Core Processor
runtime: node 22.11.0 (x64-linux)

benchmark                   avg (min … max) p75   p99    (min … top 1%)
------------------------------------------- -------------------------------
_hash with rusha              12.10 µs/iter   7.25 µs █
                        (4.66 µs … 1.27 ms) 116.62 µs █▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
_hash with native            547.04 ns/iter 435.45 ns █
                      (370.08 ns … 3.61 µs)   3.58 µs █▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁

summary
  _hash with native
   22.12x faster than _hash with rusha
```

#### Bun

```sh
bun benchmarks/rusha-vs-native.mjs 
clk: ~5.04 GHz
cpu: AMD Ryzen 7 7700 8-Core Processor
runtime: bun 1.1.37 (x64-linux)

benchmark                   avg (min … max) p75   p99    (min … top 1%)
------------------------------------------- -------------------------------
_hash with rusha              10.00 µs/iter   4.96 µs  █
                        (2.60 µs … 2.78 ms)  45.85 µs ▂█▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
_hash with native            471.82 ns/iter 420.00 ns █
                    (370.00 ns … 949.79 µs)   1.99 µs █▆▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁

summary
  _hash with native
   21.19x faster than _hash with rusha
```

### Bump level

- [ ] Major
- [X] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [X] posthog-node
- [ ] posthog-react-native

### Changelog notes

- Swtich from pure javascript ([rusha](https://www.npmjs.com/package/rusha)) to native (node:crypto) sha1 implementation


marandaneto:
@dmarticus this is used only for feature flags so your call here.

kravetsone:
This package don't need to use pure js implementation of this because it target nodejs with native implement which more faster, more secure and etc

dmarticus (approved):
This looks great!  sorry for the delay in the review; I will get this merged later today and it will go out in version 4.4.0 of the PostHog-node library.

Thank you so much for your contribution!

posthog-node/src/feature-flags.ts
```diff
@@ -458,8 +458,7 @@ class FeatureFlagsPoller {
 // # uniformly distributed between 0 and 1, so if we want to show this feature to 20% of traffic
 // # we can do _hash(key, distinct_id) < 0.2
 function _hash(key: string, distinctId: string, salt: string = ''): number {
-  // rusha is a fast sha1 implementation in pure javascript
-  const sha1Hash = createHash()
+  const sha1Hash = createHash('sha1')
   sha1Hash.update(`${key}.${distinctId}${salt}`)
   return parseInt(sha1Hash.digest('hex').slice(0, 15), 16) / LONG_SCALE
 }
```
jokull, line 464: - `crypto.subtle` is available from Node.js 17 onwards and provides a modern, native implementation for cryptographic operations.
- Switching to `crypto.subtle` can offer performance and compatibility benefits, aligning better with recent Node.js versions.

```typescript
// # This function takes a distinct_id and a feature flag key and returns a float between 0 and 1.
// # Given the same distinct_id and key, it'll always return the same float. These floats are
// # uniformly distributed between 0 and 1, so if we want to show this feature to 20% of traffic
// # we can do _hash(key, distinct_id) < 0.2
async function _hash(key: string, distinctId: string, salt: string = ''): Promise<number> {
  const data = new TextEncoder().encode(`${key}.${distinctId}${salt}`)
  const hashBuffer = await crypto.subtle.digest('SHA-1', data)
  const hashArray = Array.from(new Uint8Array(hashBuffer))
  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('')
  return parseInt(hashHex.slice(0, 15), 16) / LONG_SCALE
}
```

Issue #110, opened 2023-08-28
Posthog Survey on react native


georgibakken:
Hello! 

Not sure if this is the correct place to ask, and haven't find anything in the documentation - is [Posthog Surveys](https://posthog.com/docs/surveys/manual) available for react native ? If it is, how can it be enabled and if it's not, is there any plans of making it?

Thank you!

- [ ] [Android](https://github.com/PostHog/posthog-android/issues/102)
- [ ] [iOS](https://github.com/PostHog/posthog-ios/issues/168)
- [ ] [RN](https://github.com/PostHog/posthog-js-lite/issues/110)
- [ ] [Flutter](https://github.com/PostHog/posthog-flutter/pull/123)

[Meta issue](https://github.com/PostHog/posthog/issues/25085)

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

itsramiel:
Need this too!

marandaneto:
Hi, it's not supported yet for RN, that's correct.

Would you expect the UI (answering UI) implemented by the SDKs as well or just the API to load and send the survey answers back?

Not planned for the next few weeks at least, but we'll keep this issue open to gather interest.

itsramiel:
@marandaneto  just the API would be sufficient!

MuhammadAbdullah54321:
Support for React Native would be much appreciated.

marandaneto:
@neilkakkar is there an OpenAPI spec for surveys already or something documented people can look at?

neilkakkar:
Yep: https://posthog.com/docs/api/surveys

https://posthog.com/docs/surveys/implementing-custom-surveys

and you can check the posthog-js implementation to see how we use the APIs above: https://github.com/PostHog/posthog-js/blob/master/src/posthog-surveys.ts

marandaneto:
Thanks @neilkakkar 

If anyone reading this would like to contribute and open a PR, happy to guide and review.
It'd be pretty a port with some changes from the [JS SDK](https://github.com/PostHog/posthog-js).

andrewchester1:
It would be super cool if the react native SDK could support the Popup UI! Of course the API solution works, but I would presonally prefer the prebuild Popover in react-native over building a custom UI and handling the events myself

parthegoras-g:
Here for this too - any updates on when you expect RN support for Surveys pre-built UI (non-API/custom) to be live?

[...]

PR #333, opened 2025-01-04
Add survey support for React Native

ian-craig:
## Problem

#110 posthog-react-native doesn't have built in support for surveys

## Changes

This PR ports most of the survey logic from posthog-js into react native.

I've re-implemented most of the state handling and persistence to make it more appropriate for React Native and built on top of posthog-react-native.

See surveys/Readme.md for details on the hooks and config options I've added. I was originally going to post this as a separate library to start which is why I've split out `<PostHogSurveyProvider>` etc., but we can probably merge some of that in to the main provider.

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

Should all be backwards compatible, with the possible exception of adding a peer dependency for SVGs. We can look into what the best approach for this is.

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- TODO


marandaneto:
@ian-craig thanks a lot.
just back from the holidays, I'll catch up with a few things before and I should be able to test it/review it by EoW, thanks for doing this!

ian-craig:
@marandaneto any chance you've had some time to take a look at this? Thanks!

marandaneto:
> @marandaneto any chance you've had some time to take a look at this? Thanks!

hello, yes, I started looking into it already, expect a thoughtful review today or Monday latest =) sorry the delay!

posthog-react-native/src/posthog-rn.ts
```diff
@@ -451,4 +452,29 @@ export class PostHog extends PostHogCore {
     this.setPersistedProperty(PostHogPersistedProperty.InstalledAppBuild, appBuild)
     this.setPersistedProperty(PostHogPersistedProperty.InstalledAppVersion, appVersion)
   }
+
+  /**
+   * @todo Where should this go, and should the result be cached?
+   * I can't find a public method in PostHog which does an authenticated fetch,
+   * so for now I've included the fetch here so it can access the api key.
+   */
```
marandaneto, line 460: https://github.com/PostHog/posthog-js/blob/8d70a9ba4383b453ade94b71d7f45df4af2c892c/src/posthog-core.ts#L1294C25-L1321
i think those methods can be in the core package, [here](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-core/src/index.ts) and reuse the fetch infrastructure from there, example [here](https://github.com/PostHog/posthog-js-lite/blob/a36099a183f3a1dbdd7a68a8d5c98d39e93019ae/posthog-core/src/index.ts#L294-L323).

posthog-react-native/src/posthog-rn.ts
```diff
@@ -451,4 +452,29 @@ export class PostHog extends PostHogCore {
...
+  public async fetchSurveys(): Promise<Survey[]> {
+    const response = await this.fetch(
+      // PostHog Dashboard complains I'm using an old SDK version, I think because the ver query param isn't provided.
+      // TODO I've pulled a recent version from posthog-js, but this should be updated.
+      `${this.host}/api/surveys/?token=${this.apiKey}&ver=1.200.0`,
```
marandaneto, line 465: we can remove the version for now since the user-agent also has it https://github.com/PostHog/posthog-js-lite/blob/a36099a183f3a1dbdd7a68a8d5c98d39e93019ae/posthog-core/src/index.ts#L563

marandaneto: by doing [this](https://github.com/PostHog/posthog-js-lite/pull/333/files#r1922625072), the user agent is automatically added

posthog-react-native/src/posthog-rn.ts
```diff
@@ -451,4 +452,29 @@ export class PostHog extends PostHogCore {
...
+      {
+        method: 'GET',
+        headers: this.getCustomHeaders(),
+      }
+    )
+
+    if (response.status > 300) {
+      // TODO Best practice for handling this?
```
marandaneto, line 473: by doing [this](https://github.com/PostHog/posthog-js-lite/pull/333/files#r1922625072), error handling is builtin

posthog-react-native/src/surveys/components/SurveyModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    </Modal>
+  )
+}
+
+const styles = StyleSheet.create({
+  modalContainer: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'flex-end',
+    marginVertical: 40, // TODO Plus safe area?
```
marandaneto, line 90: makes sense

posthog-react-native/src/surveys/useActivatedSurveys.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { hasEvents } from './surveys-utils'
+import { Survey } from './posthog-surveys-types'
+import { PostHog } from '../posthog-rn'
+
+const SURVEY_SHOWN_EVENT_NAME = 'survey shown'
+
+export function useActivatedSurveys(posthog: PostHog, surveys: Survey[]): ReadonlySet<string> {
+  const [activatedSurveys, setActivatedSurveys] = useState<ReadonlySet<string>>(new Set())
+  //TODO Support Action activation
```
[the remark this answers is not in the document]
ian-craig: Makes sense, I'll leave this off

posthog-react-native/src/surveys/PostHogSurveyProvider.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  // Load surveys once
+  useEffect(() => {
+    posthog
+      .fetchSurveys()
+      .then(setSurveys)
+      .catch((error: unknown) => {
+        posthog.capture('PostHogSurveyProvider failed to fetch surveys', { error })
+      })
+  }, [posthog])
```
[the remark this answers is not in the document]
ian-craig: I've moved the fetching of surveys to Core. If I understand this comment correctly it sounds like it'd be best to check this remote config, and possibly cache the fetched surveys to prevent re-fetch, in core as well later on.

I think this `surveys` state would still be needed though as it's allowing everything else in this provider to update once the surveys finish loading.

posthog-react-native/src/surveys/components/SurveyModal.tsx
ian-craig: Okay I've added an "Optional" dependency on safe area insets like you've done with other libraries, but I'm not sure how I feel about it in this case.

The trouble is we're also dependent on the developer having a `<SafeAreaProvider>` outside where they put the `<PostHogProvider>`. I've put a fallback to some default insets if not, but it still seems like a hidden gotcha.

posthog-react-native/src/posthog-rn.ts
```diff
@@ -22,6 +22,7 @@ import {
 } from './types'
 import { withReactNativeNavigation } from './frameworks/wix-navigation'
 import { OptionalReactNativeSessionReplay } from './optional/OptionalSessionReplay'
+import { Survey, SurveyResponse } from '../../posthog-core/src/posthog-surveys-types'
```
ian-craig, line 25: TODO Remove this import now the fetch method has moved to core

posthog-react-native/src/surveys/PostHogSurveyProvider.tsx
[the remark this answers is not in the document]
marandaneto: yes, i will do the remote config next week so we can cherry pick from my changes

posthog-react-native/src/surveys/components/SurveyModal.tsx
marandaneto: we can document/add to code snippets, most people will just copy-pasta anyway

posthog-react-native/package.json
```diff
@@ -33,6 +33,7 @@
     "react-native": "^0.69.1",
     "react-native-device-info": "^10.3.0",
     "react-native-navigation": "^6.0.0",
+    "react-native-svg": "^15.2.0",
     "posthog-react-native-session-replay": "^0.1"
```
ian-craig, line 37: TODO add peer dependency for safe area context

PR #335, opened 2025-01-10
Set initial currentSessionId, log only with debug flag on.

git-lucky:
## Problem

We are seeing constant logs about the replay session Id not being rotated and it appeared to be a something going wrong. In reality it seems that this was just meant to be debug info so I wrapped it in that flag. Upon closer inspection it seems that the currentSessionId will never have a value so this is now being set after the session replay starts with a session Id the first time.

## Changes

1. Now setting a value to currentSessionId after session replay starts with one the first time.
2. Wrapped the log so its debug only.
 
### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- Fixed currentSessionId initialization and logging


marandaneto (approved):
That makes sense thanks @git-lucky

Issue #339, opened 2025-01-15
Add posthog-core to Rollup config


harrison-carter:
## Describe the solution you'd like

To use `posthog-js-lite` as a dependency within our project, it would be helpful to be able to import `posthog-core` directly from this NPM package.

Importing the `posthog-core` package to be used in a Shopify WebPixel is the use case
https://posthog.com/docs/libraries/shopify

## Describe alternatives you've considered

I am currently copying the code directly from the repo which isn't useful for dependency / version release management in order to use the PostHogCore 

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

harrison-carter:
Raised as a PR 
https://github.com/PostHog/posthog-js-lite/pull/340

marandaneto:
https://github.com/PostHog/posthog-js-lite/pull/340#issuecomment-2654989716

PR #340, opened 2025-01-15
build: add posthog core to rollup

harrison-carter:
## Problem

The PostHogCore package is not available when using `posthog-js-lite` as a dependency.

## Changes

The `posthog-core` package is added to the Rollup config

### Bump level

- [ ] Major
- [X] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native
- [X] posthog-core

### Changelog notes

- Added `posthog-core` as an exported package

### DIFF
--- a/posthog-core/index.ts
+++ b/posthog-core/index.ts
@@ -1,0 +1,1 @@
+export * from './src/index'
--- a/posthog-core/package.json
+++ b/posthog-core/package.json
@@ -1,7 +1,9 @@
 {
   "name": "posthog-core",
   "version": "2.1.0",
-  "main": "src/index.ts",
+  "main": "lib/index.cjs.js",
+  "module": "lib/index.esm.js",
+  "types": "lib/index.d.ts",
   "repository": {
     "type": "git",
     "url": "https://github.com/PostHog/posthog-js-lite.git",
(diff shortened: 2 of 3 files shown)


gazjones00:
@skoob13 is this something we could look to get merged once we've resolved the conflicts, please?

skoob13:
@harrison-carter @gazjones00 just wondering, what is the use case for the `@posthog/core` package?

gazjones00:
> @harrison-carter @gazjones00 just wondering, what is the use case for the `@posthog/core` package?

@skoob13 

Essentially, we're using `posthog-js-lite/posthog-core` in our Shopify Web Pixel to send `Purchase` events from our Shopify stores on `checkout_complete`. Due to the strict context where the Web Pixels are run, there is no direct access to the `window` object. Hence, we have to rely on the API provided by the Shopify Web Pixel context ref: https://shopify.dev/docs/api/web-pixels-api#app-web-pixels

See https://github.com/PostHog/posthog-js-lite/issues/339 for reference.

harrison-carter:
@skoob13 Would you prefer me to fix the merge conflict?

pauldambra:
can you share what an example usage of this would be?

once we expose this people will bind to it and then things are harder to change

i don't think we should expose this and should instead fix the problem by exposing a windowless something or fixing whichever public package you couldn't use

pauldambra (changes requested):
this feels too broad to me, we should fix the problem as in my other comment

harrison-carter:
We've explored a different avenue using the Node package to fit the use case.

Thanks for the response's thus far 👍🏼

PR #351, opened 2025-01-21
Fix: Vercel AI + Fix

k11kirky:
https://www.loom.com/share/a5a02c23073943559b204ecfcb3e11cc


skoob13:
The approach seems reasonable. Version 3.4.0 introduced OpenTelemetry tracing. According to the NPM, version 3.2 is still the most popular version with 340k downloads. I would go with this approach as it's the most seamless and does not require version updates.

posthog-ai/src/vercel/middleware.ts
```diff
@@ -0,0 +1,157 @@
...
+                });
+                throw error;
+            }
+        },
+    };
+
+    return middleware;
+};
+
+export const wrapVercelLanguageModel = (
```
skoob13, line 139: Since we already import this decorator from @posthog/posthog-a/vercel, can we simplify it? Like `withTracing`.

posthog-ai/package.json
```diff
@@ -27,8 +27,10 @@
     "llm"
   ],
   "dependencies": {
+    "ai": "^4.1.0",
```
skoob13, line 30: I'd make both ai and openai peer dependencies. Is there any reason to keep them in dependencies? Under peer deps, the user is in control of which version they want to use.

posthog-ai/src/vercel/middleware.ts
```diff
@@ -0,0 +1,162 @@
...
+import { v4 as uuidv4 } from 'uuid'
+import type { PostHog } from 'posthog-node'
+import { sendEventToPosthog } from '../utils'
+
+interface CreateInstrumentationMiddlewareOptions {
+  posthog_distinct_id?: string
+  posthog_trace_id: string
+  posthog_properties?: Record<string, any>
+  posthog_privacy_mode?: boolean
+  posthog_groups?: string[]
```
skoob13, line 16: We should use a camel case for parameters. I think the packages in this repo all use camel case.

skoob13 (approved):
Nothing blocking. Merge it if it's urgent.

PR #361, opened 2025-01-24
feat: error handling and azure

k11kirky:
Added field $ai_is_error as a bool and stringified the error into $ai_error so we can show errors in the ui
Adde AzureOpenAI to the package


skoob13 (approved):
All good

PR #369, opened 2025-02-03
fix(session-replay): improve session replay linked flag type handling

skoob13:
## Problem

Noticed errors in the production version of a React Native app, complaining of the line with a typecast. When the config is null, it goes through the `typeof ... === 'object'` condition, as `typeof` for null returns `object`.

## Changes

Fix the line to check for null values.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

### Changelog notes

- fix: improve session replay linked flag type handling

### DIFF
--- a/posthog-react-native/CHANGELOG.md
+++ b/posthog-react-native/CHANGELOG.md
@@ -27,7 +27,7 @@
 # 3.4.0 - 2024-11-26
 
 1. feat: automatically mask out user photos and sandboxed views like photo picker (iOS Only)
-1. To disable masking set `maskAllSandboxedViews` and `maskPhotoLibraryImages` to false
+   1. To disable masking set `maskAllSandboxedViews` and `maskPhotoLibraryImages` to false
 
 ```js
 export const posthog = new PostHog(
--- a/posthog-react-native/CHANGELOG.md
+++ b/posthog-react-native/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Next
 
+# 3.6.4 - 2025-02-03
+
+1. fix: improve session replay linked flag type handling
+
 # 3.6.3 - 2025-01-16
 
 1. fix: session replay respect linked feature flags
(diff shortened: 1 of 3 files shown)


posthog-react-native/src/posthog-rn.ts
```diff
@@ -336,15 +336,15 @@ export class PostHog extends PostHogCore {
...
+    const linkedFlag = decideReplayConfig['linkedFlag'] as string | { [key: string]: JsonType } | null | undefined
     if (typeof linkedFlag === 'string') {
       const value = decideFeatureFlags[linkedFlag]
       if (typeof value === 'boolean') {
         recordingActive = value
       }
 
       this.logMsgIfDebug(() => console.log('PostHog Debug', `Session replay ${linkedFlag} linked flag value: ${value}`))
-    } else if (typeof linkedFlag === 'object') {
+    } else if (linkedFlag && typeof linkedFlag === 'object') {
```
marandaneto, line 347: L340 will need this check too or we move both `if`s inside of an `if (linkedFlag)` check

posthog-react-native/CHANGELOG.md
```diff
@@ -23,7 +27,7 @@
 # 3.4.0 - 2024-11-26
 
 1. feat: automatically mask out user photos and sandboxed views like photo picker (iOS Only)
-  1. To disable masking set `maskAllSandboxedViews` and `maskPhotoLibraryImages` to false
+1. To disable masking set `maskAllSandboxedViews` and `maskPhotoLibraryImages` to false
```
marandaneto, line 30:
```suggestion
   1. To disable masking set `maskAllSandboxedViews` and `maskPhotoLibraryImages` to false
```

marandaneto:
Thanks @skoob13 
left a comment!

posthog-react-native/src/posthog-rn.ts
skoob13: I'm not quite following. Why does the line 340 need the same check?

marandaneto: L340 also does `if (typeof linkedFlag === 'string') {` so its the same issue.
ah nvm there we compare `linkedFlag` with `string` and not `object`

marandaneto (approved):

PR #371, opened 2025-02-04
feat: langchain support

k11kirky:


posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+    }
+    if (metadata) {
+      if (metadata.ls_model_name) {
+        generation.model = metadata.ls_model_name
+      }
+      if (metadata.ls_provider) {
+        generation.provider = metadata.ls_provider
+      }
+    }
+    if (serialized && 'kwargs' in serialized && serialized.kwargs.openai_api_base) {
```
skoob13, line 298: Surprised to see kwargs in JS, but indeed the SDK [has](https://github.com/langchain-ai/langchainjs/blob/458790627318c9b15ea1c34c22b0b9774d7f8e98/langchain-core/src/load/serializable.ts#L13) it.

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+  public name = 'PosthogCallbackHandler'
+  private client: PostHog
+  private distinctId?: string | number
+  private traceId?: string | number
+  private properties: Record<string, any>
+  private privacyMode: boolean
+  private groups: Record<string, any>
+  private debug: boolean
+
+  private runs: RunMetadataStorage = {}
```
skoob13, line 49: Nit: I'd use a Map for storage

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+  private parentTree: { [runId: string]: string } = {}
```
skoob13, line 50: Nit: Map also makes more sense here

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+  // ===== PRIVATE HELPERS =====
+
+  private _setParentOfRun(runId: string, parentRunId?: string): void {
+    if (parentRunId) {
+      this.parentTree[runId] = parentRunId
+    }
+  }
+
+  private _popParentOfRun(runId: string): void {
+    delete this.parentTree[runId]
```
skoob13, line 245: Returning to the nit with Map(), delete is generally slower than the delete method of Map.

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+  private _popRunAndCaptureTraceOrSpan(
+    runId: string,
+    parentRunId: string | undefined,
+    outputs: ChainValues | DocumentInterface[] | AgentFinish | Error | any
+  ): void {
+    const traceId = this._getTraceId(runId)
+    this._popParentOfRun(runId)
+    const run = this._popRunMetadata(runId)
+    if (!run) return
+    if ('modelParams' in run) {
```
skoob13, line 337: It would be slightly more verbose to implement a typeguard function for the run value

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+      eventProperties['$process_person_profile'] = false
+    }
+    if (outputs instanceof Error) {
+      eventProperties['$ai_error'] = outputs.toString()
+      eventProperties['$ai_is_error'] = true
+    } else if (outputs !== undefined) {
+      eventProperties['$ai_output_state'] = withPrivacyMode(this.client, this.privacyMode, outputs)
+    }
+    this.client.capture({
+      distinctId: (this.distinctId as string) || runId,
```
skoob13, line 376:
```suggestion
      distinctId:  this.distinctId ? String(this.distinctId) : runId,
```

Two things:
- `distinctId` can be theoretically 0, which we won't capture.
- `distinctId` by the type is a string. A number might break things.

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+      }
+    }
+
+    Object.assign(eventProperties, this.properties)
+    if (!this.distinctId) {
+      eventProperties['$process_person_profile'] = false
+    }
+
+    this.client.capture({
+      distinctId: (this.distinctId as string) || traceId,
```
skoob13, line 453:
```suggestion
      distinctId:  this.distinctId ? String(this.distinctId) : runId,
```

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+
+  private _popRunAndCaptureGeneration(
+    runId: string,
+    parentRunId: string | undefined,
+    response: LLMResult | Error
+  ): void {
+    const traceId = this._getTraceId(runId)
+    this._popParentOfRun(runId)
+    const run = this._popRunMetadata(runId)
+    if (!run || !('modelParams' in run)) {
```
skoob13, line 391: Also feels like a type guard needed

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+  private _logDebugEvent(eventName: string, runId: string, parentRunId: string | undefined, extra: any): void {
+    if (this.debug) {
+      console.log(`Event: ${eventName}, runId: ${runId}, parentRunId: ${parentRunId}, extra:`, extra)
+    }
+  }
+
+  private _getLangchainRunName(serialized: any, ...args: any[]): string | undefined {
+    if (args && args.length > 0) {
+      for (const arg of args) {
+        if (arg && arg.name) return arg.name
```
skoob13, line 469: It might be better actually to check if the arg is an object like:
```suggestion
        if (arg && typeof arg === 'object' && 'name' in arg && typeof arg.name === 'string') {
          return arg.name
        }
```
It might be an overkill, though.

skoob13 (changes requested):
`distinctId` serialization is blocking, but anything else looks good

posthog-ai/src/langchain/callbacks.ts
```diff
@@ -0,0 +1,578 @@
...
+  // ===== CALLBACK METHODS =====
+
+  public handleChainStart(
+    chain: Serialized,
+    inputs: ChainValues,
+    runId: string,
+    parentRunId?: string,
+    tags?: string[],
+    metadata?: Record<string, unknown>,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
skoob13, line 83: You can omit to specify these params btw

skoob13 (approved):

Issue #286, opened 2024-10-14
Set locale and timezone info if not using Expo


marandaneto:
### Description

https://github.com/zoontek/react-native-localize
https://github.com/PostHog/posthog-js-lite/blob/2b56d75c1c3b0f3c67bbec7a4aeed25ab2c52399/posthog-react-native/src/native-deps.tsx#L38-L41

This is only set if using Expo but not other alternative is given

PR #373, opened 2025-02-05
chore: set locale and timezone using the react-native-localize library

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js-lite/issues/286

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-ai/src/anthropic/index.ts
+++ b/posthog-ai/src/anthropic/index.ts
@@ -72,7 +72,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
       return parentPromise.then((value) => {
         const passThroughStream = new PassThrough({ objectMode: true })
         let accumulatedContent = ''
-        let usage: { inputTokens: number; outputTokens: number } = {
+        const usage: { inputTokens: number; outputTokens: number } = {
           inputTokens: 0,
           outputTokens: 0,
         }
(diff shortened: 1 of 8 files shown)


marandaneto:
@k11kirky tagged you because `yarn install` updated a few things, the linter too, just double check its ok

posthog-ai/src/anthropic/index.ts
```diff
@@ -72,7 +72,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
       return parentPromise.then((value) => {
         const passThroughStream = new PassThrough({ objectMode: true })
         let accumulatedContent = ''
-        let usage: { inputTokens: number; outputTokens: number } = {
+        const usage: { inputTokens: number; outputTokens: number } = {
```
k11kirky, line 75: this should have stayed let - its gets resigned under

k11kirky (approved):

posthog-ai/src/anthropic/index.ts
marandaneto: the linter did it automatically btw (eslint, prettier), but as I can see it, the usage props are resigned, but not the `usage` itself.
Happy to change it back but am I missing something?

PR #384, opened 2025-02-11
Deprecate `is_simple_flag`

haacked:
And remove it from tests.

This is a follow-up to https://github.com/PostHog/posthog-python/pull/186. See that issue for more context.


marandaneto:
nice, @haacked a few python and TS files that still use the is_simple_flag prop https://github.com/search?q=repo%3APostHog%2Fposthog+is_simple_flag&type=code
should all this code be removed as well?

marandaneto (approved):

PR #385, opened 2025-02-11
chore: cache surveys in storage

marandaneto:
### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for X


marandaneto:
> code looks good! one question though:
> 
> 1. when are the local persisted surveys invalidated? i.e. we make the network call again. is this every time we init the posthog sdk?

The storage is isolated per apiKey, it's gonna be reset if the user, eg calls `posthog.reset` or when a new request is made with new results (every time the SDK inits, yes)

PR #386, opened 2025-02-11
fix:Specify tz with the date

haacked:
This test was failing on my local machine because of TZ differences. Here's the breakdown.

1. System Date: `2022-05-01`. Jest interprets it as `2022-05-01 00:00:00 UTC`
2. 1hr before `2022-04-30 23:00:00 UTC`
3. Override value: `2022-04-30 22:00:00` which JavaScript interprets as PDT (I live on the west coast of the U.S.). Converted to UTC -> `2022-05-01 05:00:00 UTC`. That's definitely not before `2022-05-01 00:00:00 UTC` and the test fails.

Fix: Specify GMT for the override value.

Some of the other tests might break for other TZ. We should probably mock the TZ in the tests to be more explicit. But I just want to make these tests pass my machine for now.

### DIFF
--- a/posthog-node/test/feature-flags.spec.ts
+++ b/posthog-node/test/feature-flags.spec.ts
@@ -2038,7 +2038,7 @@ describe('match properties', () => {
     ['is_date_after', '1h', '2022-04-30', false],
     // # Try all possible relative dates
     ['is_date_before', '1h', '2022-05-01 00:00:00', false],
-    ['is_date_before', '1h', '2022-04-30 22:00:00', true],
+    ['is_date_before', '1h', '2022-04-30 22:00:00 GMT', true],
     ['is_date_before', '-1d', '2022-04-29 23:59:00 GMT', true],
     ['is_date_before', '-1d', '2022-04-30 00:00:01 GMT', false],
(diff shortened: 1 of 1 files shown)


posthog-node/test/feature-flags.spec.ts
```diff
@@ -2038,7 +2038,7 @@ describe('match properties', () => {
     ['is_date_after', '1h', '2022-04-30', false],
     // # Try all possible relative dates
     ['is_date_before', '1h', '2022-05-01 00:00:00', false],
```
rafaeelaudibert, line 2040: Can you add GMT here as well for consistency? I agree the right fix is to mock the timezone during tests and that it can be done some other time

rafaeelaudibert (approved):

posthog-node/test/feature-flags.spec.ts
haacked: No problem! I wanted to fix all the dates, but blanketly adding `GMT` is not the correct fix. I'd have to do the date analysis on each of them. So I punted the rest of the dates for now. 😛

PR #387, opened 2025-02-12
Do not enrich the $feature_flag_called event with more feature flag data

haacked:
See https://github.com/PostHog/posthog-python/pull/181 for more context. This PR is a port of that fix.

## Problem

The `client.getFeatureFlag()` method will by default call `client.capture` with the event named `$feature_flag_called`. 

Some time ago, the `capture()` method was changed to enrich the event with all locally cached feature flags even if `sendFeatureFlags` is `false`.

What this leads to is `client.getFeatureFlag()` -> `client.capture()` -> `client.getAllFlags()`. But the call to `client.getAllFlags()` isn't passed the original `person_properties` that were passed to `client.getFeatureFlag()`. This leads to inconsistency in the enriched data.

This PR takes the simplest path by not enriching the captured event with all feature flags for the `$feature_flag_called` event.

## Changes

If the captured event is `$feature_flag_called`, we don't enrich it with all flags.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Fixed edge case where `$feature_flag_called` events were enriched with additional feature flag data when they shouldn't be.

### DIFF
--- a/posthog-node/CHANGELOG.md
+++ b/posthog-node/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Next
 
-# 4.5.1 - 2025-02-11
+# 4.5.1 - 2025-02-12
 
 1. fix: Fixed edge case where `$feature_flag_called` events were enriched with additional feature flag data when they shouldn't be.
 
--- a/posthog-node/CHANGELOG.md
+++ b/posthog-node/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Next
 
+# 4.5.1 - 2025-02-11
+
+1. fix: Fixed edge case where `$feature_flag_called` events were enriched with additional feature flag data when they shouldn't be.
+
 # 4.5.0 - 2025-02-06
 
 ## Added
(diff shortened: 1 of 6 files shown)


posthog-node/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Next
 
+# 4.5.1 - 2025-02-11
```
marandaneto, line 3:
```suggestion
# 4.5.1 - 2025-02-12
```

marandaneto:
@haacked all mobile SDKs do the same iirc.
I wonder if this fix should be ported everywhere since on mobile, flags are cached locally due to the device having no connectivity etc and flags might be outdated when reading flags/capturing the feature flag called

posthog-node/CHANGELOG.md
haacked: Ha! This is a moving target. 😄

haacked:
> I wonder if this fix should be ported everywhere since on mobile, flags are cached locally due to the device having no connectivity etc and flags might be outdated when reading flags/capturing the feature flag called

Indeed, I think it should be ported to anywhere we have this specific issue.

PR #388, opened 2025-02-12
feat(flags): Add method for fetching decrypted remote config flag payload

havenbarnes:
## Problem

Context: [PostHog/posthog#27414 (files)](https://github.com/PostHog/posthog/pull/27414/files)

Screenshot 2025-02-11 at 5 07 21 PM

## Changes

Adds new `getDecryptedFeatureFlagPayload` method to Node SDK

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### DIFF
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -452,8 +452,8 @@ class FeatureFlagsPoller {
     clearTimeout(this.poller)
   }
 
-  _requestDecryptedFeatureFlagPayload(flagId: number): Promise<PostHogFetchResponse> {
-    const url = `${this.host}/api/projects/@current/feature_flags/${flagId}/remote_config/`
+  _requestDecryptedFeatureFlagPayload(flagKey: string): Promise<PostHogFetchResponse> {
+    const url = `${this.host}/api/projects/@current/feature_flags/${flagKey}/remote_config/`
 
     const options: PostHogFetchOptions = {
       method: 'GET',
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -456,7 +456,7 @@ class FeatureFlagsPoller {
     clearTimeout(this.poller)
   }
 
-  _requestDecryptedFeatureFlagPayload(flagKey: string): Promise<PostHogFetchResponse> {
+  _requestRemoteConfigPayload(flagKey: string): Promise<PostHogFetchResponse> {
     const url = `${this.host}/api/projects/@current/feature_flags/${flagKey}/remote_config/`
 
     const options = this.getPersonalApiKeyRequestOptions()
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -461,7 +461,19 @@ class FeatureFlagsPoller {
 
     const options = this.getPersonalApiKeyRequestOptions()
 
-    return this.fetch(url, options)
+    let abortTimeout = null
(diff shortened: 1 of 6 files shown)


havenbarnes:
Hey @rafaeelaudibert  or maybe @marandaneto , what's the latest on version bumping / changelog, should I do that manually as it is currently or is it automated?

posthog-node/src/feature-flags.ts
```diff
@@ -451,6 +451,21 @@ class FeatureFlagsPoller {
   stopPoller(): void {
     clearTimeout(this.poller)
   }
+
+  _requestDecryptedFeatureFlagPayload(flagId: number): Promise<PostHogFetchResponse> {
+    const url = `${this.host}/api/projects/@current/feature_flags/${flagId}/remote_config/`
```
haacked, line 456: What is `@current`? Is that a literal part of the URL? Or does that get replaced later?

haacked (approved):
LGTM! Just one question, but I don't see anything blocking.

rafaeelaudibert:
> Hey @rafaeelaudibert or maybe @marandaneto , what's the latest on version bumping / changelog, should I do that manually as it is currently or is it automated?

I haven't played with the js-lite repo yet, so I'm afraid I don't know. 

I see that the `bump minor` label exists in this repo, so I say give it a go and let's see hahah

dmarticus (approved):

posthog-node/src/feature-flags.ts
marandaneto: I also wondered about this when reviewing the other PR
https://github.com/PostHog/posthog/blob/01acee6e26715f739294e036752ea21551526baf/frontend/src/toolbar/flags/flagsToolbarLogic.ts#L55
So it is part of the URL, but I have no idea how the server decides which project it is about (maybe there's a default project settings)?

posthog-node/src/feature-flags.ts
```diff
@@ -451,6 +451,21 @@ class FeatureFlagsPoller {
...
+
+    const options: PostHogFetchOptions = {
+      method: 'GET',
+      headers: {
+        ...this.customHeaders,
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${this.personalApiKey}`,
+      },
+    }
```
marandaneto, line 465: nit: since this is defined elsewhere, could we extract it to a private function?

posthog-node/src/feature-flags.ts
```diff
@@ -451,6 +451,21 @@ class FeatureFlagsPoller {
...
+
+    return this.fetch(url, options)
```
marandaneto, line 467: I think a try/catch would be useful either here or in the caller

posthog-node/src/feature-flags.ts
havenbarnes: The django app is smart enough to infer `@current` from the token used to make the request! I didn't want to require the user to find their project ID just to use this feature

haacked: > The django app is smart enough to infer `@current` from the token used to make the request! I didn't want to require the user to find their project ID just to use this feature

Oh cool! I didn't know that.

posthog-node/src/feature-flags.ts
havenbarnes: Done

Issue #391, opened 2025-02-13
Feature flags permanently undefined after signing out/posthog.reset()


otajor:
## Bug description

We have a react native app with expo. We are seeing odd behaviour where feature flags that are enabled on the dashboard work fine on app load before login, work fine after logging in (when we call posthog.identify), but then if the user logs out (when we call posthog.reset), the feature flag stops working and isEnabled returns undefined. We are using the FF hooks e.g. 
```
const [isEnabled, payload] = useFeatureFlagWithPayload('show-modal-with-image');
```
or
```
const featureXFlagValue = useFeatureFlag('featureX');
```
Instead of the expected boolean value, all return values are undefined so the flags don't work. This remains the case **even if the user signs in again** (so we call posthog.identify again). The feature flags still return undefined, until we reload the app, at which point the problem is resolved until the user signs out again.

## How to reproduce

1. Have a feature flag enabled and set up with one of the above hooks, check that it works and returns boolean values (true if flag is enabled)
2. Call posthog.identify() with user data
3. Call posthog.reset()
4. Check what values the feature flag is returning.

Expected: boolean values
Actual: undefined

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [X] posthog-react-native

marandaneto:
Hello @otajor 
I think [this](https://github.com/PostHog/posthog-js-lite/pull/392) should fix it.
A workaround for the current version is to call `reset`, `await reloadFeatureFlagsAsync`, and `identify` again.

`reset` will clear all loaded feature flags since it's now a different user.
Once `identify` is called, it will load feature flags again but be aware that the request is async so it takes a couple millis until the flags are available again.

Let us know if the new version works once it's merged and published.

otajor:
@marandaneto wow speedy work! Thanks, calling `reloadFeatureFlags` after logging out solves our problem for now. We'll test the fix once it's released - thanks again 🙏

PR #392, opened 2025-02-13
fix: ensure feature flags are reloaded after reset() to prevent undefined values

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js-lite/issues/391

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -1195,7 +1195,6 @@ export abstract class PostHogCore extends PostHogCoreStateless {
               newFeatureFlags = { ...currentFlags, ...res.featureFlags }
               newFeatureFlagPayloads = { ...currentFlagPayloads, ...res.featureFlagPayloads }
             }
-
             this.setKnownFeatureFlags(newFeatureFlags)
             this.setKnownFeatureFlagPayloads(
               Object.fromEntries(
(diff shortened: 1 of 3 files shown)


posthog-core/src/index.ts
```diff
@@ -1148,7 +1150,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
   private async decideAsync(sendAnonDistinctId: boolean = true): Promise<PostHogDecideResponse | undefined> {
     await this._initPromise
     if (this._decideResponsePromise) {
-      return this._decideResponsePromise
+      await this._decideResponsePromise
```
marandaneto, line 1153: if reload flags were called multiple times (eg reset, reload, identify, etc), the requests should be executed one after the other with the new values (eg new distinct id, etc) rather than returning the current  request maybe with the wrong request values

posthog-core/src/index.ts
```diff
@@ -835,6 +835,8 @@ export abstract class PostHogCore extends PostHogCoreStateless {
           this.setPersistedProperty((PostHogPersistedProperty as any)[key], null)
         }
       }
+
+      this.reloadFeatureFlags()
```
marandaneto, line 839: reset changes from identified user to non identified user, and all flags were cleared, so now we have to load all flags as anon user

dmarticus (approved):
ooof, nice fix!

Issue #212, opened 2024-03-26
Breaking changes for the next major


marandaneto:
### Description

- [x] Remove `version` and `build` in favor of the $ props (Also for Native SDKs - Android and iOS)
- [x] Remove `personProperties` in favor of `setPersonPropertiesForFlags`
- [x] Remove `groupProperties` in favor of `setGroupPropertiesForFlags`
- [x] Fix the confusion between `options.captureNativeAppLifecycleEvents` and `autocapture.captureLifecycleEvents`

marandaneto:
https://github.com/PostHog/posthog-js-lite/pull/520

PR #398, opened 2025-02-19
fix: Autocapture native app lifecycle events

rafaeelaudibert:
## Problem

Right now, we have two configurations for autotracking lifecycle events, and we suggest customers to only use one or another. The problem is that they're not equivalent.

The `options.captureNativeAppLifecycleEvents` option is capturing both app lifecycle events ("Application Opened", "Application Backgrounded", "Application Became Active") but also native events ("Application Installed" and "Application Updated").

`autocapture.captureLifecycleEvents` on the other hand does not capture the latter.

## Changes

Let's make them act in the same way by using the option from inside `autocapture` - the one we suggest in our docs - and copying it over to the `options` config hash. They'll now behave in the same way, there isn't even a problem setting them both at the same time - we might need to update docs.

See https://posthoghelp.zendesk.com/agent/tickets/25578

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [x] posthog-react-native

### Changelog notes

fix: Autocapture native app lifecycle events


posthog-react-native/src/hooks/useLifecycleTracker.ts
```diff
@@ -2,18 +2,24 @@ import { useEffect, useRef } from 'react'
...
+  const openTrackedRef = useRef<boolean>(false)
   const contextClient = usePostHog()
   const posthog = client || contextClient
 
   return useEffect(() => {
     if (!openTrackedRef.current) {
       openTrackedRef.current = true
-      // TODO: add missing initialUrl
-      posthog.capture('Application Opened')
+      captureApplicationOpened(posthog)
```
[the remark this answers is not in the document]
rafaeelaudibert: I don't mind if it fails

posthog-react-native/src/PostHogProvider.tsx
```diff
@@ -49,28 +43,41 @@ export const PostHogProvider = ({
...
+    const parsedOptions = {
+      ...options,
+
+      // For backwards-compatible reasons, this information can be stored in two separate places
+      // Default to true if not set
+      captureNativeAppLifecycleEvents:
+        options.captureNativeAppLifecycleEvents !== undefined
+          ? options.captureNativeAppLifecycleEvents
+          : !captureNone && (captureAll || (autocaptureOptions?.captureLifecycleEvents ?? true))
+    }
```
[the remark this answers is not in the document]
rafaeelaudibert: Nice

posthog-react-native/src/PostHogProvider.tsx
```diff
@@ -49,28 +43,41 @@ export const PostHogProvider = ({
...
+
+    return new PostHog(apiKey ?? '', parsedOptions)
+  }, [client, apiKey]) // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
rafaeelaudibert: Now you're just crazy, `autocaptureOptions` is not in the dependency array

posthog-react-native/src/hooks/useLifecycleTracker.ts
```diff
@@ -2,18 +2,24 @@ import { useEffect, useRef } from 'react'
...
+import { Linking } from 'react-native'
+
+async function captureApplicationOpened(posthog: PostHog): Promise<void> {
+  const url = (await Linking.getInitialURL()) ?? undefined
+  posthog.capture('Application Opened', { url })
+}
 
 export function useLifecycleTracker(client?: PostHog): void {
-  const openTrackedRef = useRef(false)
+  const openTrackedRef = useRef<boolean>(false)
```
marandaneto, line 13: https://github.com/PostHog/posthog-js-lite/blob/ac83f9a98806e7a4c2c32d38ea874cc31c41c645/posthog-react-native/src/frameworks/wix-navigation.ts#L20-L40
does something very similar, we could delete that function, extract and reuse the `useLifecycleTracker` in the wix navigation?

marandaneto: another issue is that `useLifecycleTracker` isn't being used anywhere?
can we actually get rid of this now (and trackLifecycle) since its all controlled via `captureNativeAppLifecycleEvents`

marandaneto:
One issue I see is that after this PR, you cannot opt out of the app install and app update events.
That is the difference between `captureNativeAppLifecycleEvents` and `captureLifecycleEvents` I think, ps: they were added past me, so I agree that it  is confusing but I don't have context on why it was done this way

posthog-react-native/src/hooks/useLifecycleTracker.ts
rafaeelaudibert: Yeah, happy to update `wix-navigation` to use that hook. The reason why I didnt remove the hook is that it's exposed, so someome might be using it directly without our knowledge, so I thought of keeping it around until the next major.

marandaneto: i think its fine because we dont document `useLifecycleTracker` anywhere so its safe to remove since we also have an alternative, same for `trackLifecycle`

rafaeelaudibert:
> One issue I see is that after this PR, you cannot opt out of the app install and app update events. That is the difference between `captureNativeAppLifecycleEvents` and `captureLifecycleEvents` I think, ps: they were added past me, so I agree that it is confusing but I don't have context on why it was done this way

Yeah, 100% true. In my mind, the best solution is to break the existing API compatibility with a major and then split this setting up in two at the config level rather than having some of it running at the provider level like we have now. One problem we have now is that you can't opt in to native lifecycle events without opting in to lifecycle events, so it's already kinds restrictive anyway.

So, we have two options: get a new major ready now, and we might wanna include more things in it, or just keep this as is and I can tell the user to move from the `autocapture` setting to the `options` setting - and then update the docs to mention how quirky this is

marandaneto:
> > One issue I see is that after this PR, you cannot opt out of the app install and app update events. That is the difference between `captureNativeAppLifecycleEvents` and `captureLifecycleEvents` I think, ps: they were added past me, so I agree that it is confusing but I don't have context on why it was done this way
> 
> Yeah, 100% true. In my mind, the best solution is to break the existing API compatibility with a major and then split this setting up in two at the config level rather than having some of it running at the provider level like we have now. One problem we have now is that you can't opt in to native lifecycle events without opting in to lifecycle events, so it's already kinds restrictive anyway.
> 
> So, we have two options: get a new major ready now, and we might wanna include more things in it, or just keep this as is and I can tell the user to move from the `autocapture` setting to the `options` setting - and then update the docs to mention how quirky this is

tbh I think this was just an oversight, the wording 'native' lifecycles does not even reflect anything 'native' just because it captures 2 more possible events.
Added to https://github.com/PostHog/posthog-js-lite/issues/212
let me think about it for a sec.

marandaneto:
@rafaeelaudibert https://github.com/PostHog/posthog-js-lite/pull/401
an idea is to just merge them, document better as I did in this PR, and on the next major, really have a single option.

I think the reason why 'native' and not 'native' is because 'native' might have the storage set up and can compare the versions (if its a new install or an update) and not native there's no storage at all (which is not true since you can set up your own storage function/its just not automatic), so it's confusing but I guess we'd be fine with a minor after merging my PR.
it'd be a minor because we'd add a feature on top of the current one.

if one does not want the new approach with app install and app update events, they can simply set up their own `AppState.addEventListener` which is just a couple of lines instead having a more granular config for each app lifecycle event.

WDYT?

rafaeelaudibert:
> @rafaeelaudibert #401 an idea is to just merge them, document better as I did in this PR, and on the next major, really have a single option.
> 
> I think the reason why 'native' and not 'native' is because 'native' might have the storage set up and can compare the versions (if its a new install or an update) and not native there's no storage at all (which is not true since you can set up your own storage function/its just not automatic), so it's confusing but I guess we'd be fine with a minor after merging my PR. it'd be a minor because we'd add a feature on top of the current one.
> 
> if one does not want the new approach with app install and app update events, they can simply set up their own `AppState.addEventListener` which is just a couple of lines instead having a more granular config for each app lifecycle event.
> 
> WDYT?

Ok, I like that, especially the fact that they can just roll their own if they want a more customizable experience. Then the next major can add specific configs for each individual kind of event - or add something akin to `posthog-js`'s `before_send` where they can filter these events out.

Just to confirm: Are you proposing we just merge your PR on this one, and then both go live as a minor? I'm happy to do that

---

On another note, I don't know how releases work on this repo, and I couldn't find any info in the README - it's too barebones IMO.

marandaneto:
@rafaeelaudibert 
> Just to confirm: Are you proposing we just merge your PR on this one, and then both go live as a minor? I'm happy to do that

yes, the release is simple, all info is [here](https://github.com/PostHog/posthog-js-lite/blob/main/RELEASING.md)

posthog-react-native/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Next
 
+1. fix: Autocapture native app lifecycle events
```
```suggestion
# 3.11.0 - 2025-02-20

1. fix: Autocapture native app lifecycle events
```

would need to change the version in the package json as well

rafaeelaudibert:
@marandaneto Just merged your branch here, will get the release ready - by EOD, I'm working on some other stuff already

marandaneto (approved):

PR #403, opened 2025-02-21
feat(feature-flags): support quota limiting for feature flags

dmarticus:
with [this PR](https://github.com/PostHog/posthog/pull/28564), we now start to respond different with the `/decide` and `/local_evaluation` APIs if users have gone over their quota limit.  Now we need to change the SDKs to handle these new responses.


posthog-node/src/feature-flags.ts
```diff
@@ -391,6 +391,17 @@ class FeatureFlagsPoller {
         )
       }
 
+      if (res && res.status === 402) {
```
marandaneto, line 394: 2 questions:
does the API already return 402, or is that a new thing? this would break back compatibility with older SDK versions, not handling errors correctly, maybe retrying too much, etc.
should it be 429 instead? 429 is usually used for rate limiting, which is not the very same as quota exceed, just food for thought

dmarticus: > does the API already return 402, or is that a new thing? this would break back compatibility with older SDK versions, not handling errors correctly, maybe retrying too much, etc.

Definitely a new thing, so older versions might not handle quota limiting as gracefully, but they'd still get quota limited – it would just be an unhandled exception that's still swallowed top-level, it just might not look as well-defined.  I don't know if there's a better way to handle compatibility in this case, if so, I feel like I'd need to add it to the API itself, and not the SDK (which is maybe what you're suggesting).  Maybe I can do some sort of thing where I add a version flag (either as a query string or a request header) to the PostHog API that the SDKs start sending with their requests, and have old versions give a different response, or something like that.

At the end of the day, though, I still want the local_evaluation endpoint to fail if we reach a quota limit, so maybe having it be an unhandled exception that's caught is fine.  What do you think?

> should it be 429 instead? 429 is usually used for rate limiting, which is not the very same as quota exceed, just food for thought

We debated this, and I prefer 402 since we're not rate-limiting the users, we're limiting them based on payment.  Plus it avoids confusion with the existing 429 error that the local_evaluation endpoint returns when we _do_ rate-limit customers.

marandaneto: > they'd still get quota limited – it would just be an unhandled exception that's still swallowed top-level

I think it really depends on the SDK implementation, are we sure 100% about all SDKs? it's a huge breaking change, IMO (including breaking clients calling the API directly).
For example, some SDKs would not read the payload at all just because an error was thrown, and the /decide also returns remote config and not only flags, which is essential in some cases.
For example, the Android and iOS SDK won't enable session recording just because flags exceed quota, but it has recordings quota.

> I don't know if there's a better way to handle compatibility in this case, if so, I feel like I'd need to add it to the API itself, and not the SDK (which is maybe what you're suggesting)

Yes, 100%, I think there are a few solutions, a header, a new body payload such as `quotaLimited`, but not a HTTP error (4xx|5xx).

> Maybe I can do some sort of thing where I add a version flag (either as a query string or a request header) to the PostHog API that the SDKs start sending with their requests, and have old versions give a different response, or something like that.

That is another alternative, eg /decide/?v=4, so v4 will return an HTTP error, or as said above, a new header, a new payload, etc.

> We debated this, and I prefer 402 since we're not rate-limiting the users, we're limiting them based on payment. Plus it avoids confusion with the existing 429 error that the local_evaluation endpoint returns when we do rate-limit customers.

That is fine, no strong opinion here, but I think its more common 403 or 429.
429 is for requests limit (More like abuse)
403 is for hard quota limit usually.
402 yeah you can pay/upgrade plan, etc
so 402 is fine, but again, I don't think returning an error is ideal until the remote config API is fully rolled out and older SDKs depend on it.

dmarticus: All your points are good, and I generally agree with the sentiment!  However, I think that we're slightly missing the point here: the change to return a `402` error whenever we quota limit a resource (like flags) is _not_ happening to the `/decide` endpoint – for all the reasons you described above, I don't want to roll a new type of error to `/decide`, and agree that decide shouldn't fail if one resource is quota-limited.  The change we're talking about is a change to the `/local_evaluation` endpoint, which is _only_ used for flag evaluation and already has an existing paradigm in all SDKs for failing with a 400 error in the event that it's (a) rate-limited or (b) has incorrect auth (two common cases). 

For `/decide`, we don't fail at all, and we actually do exactly this

> Yes, 100%, I think there are a few solutions, a header, a new body payload such as quotaLimited, but not a HTTP error (4xx|5xx).

We return a new payload called `quotaLimited` that is a list of resources that were quota limited.  Specifically, it works like this: https://github.com/PostHog/posthog/pull/28564/files#diff-614b64aa6eedf163ceb5254204e0702dc10b6dede3fdb66e9797c933ad828033R285-R298

So, overall, there's 2 situations here:

1. `/decide`, which won't fail even if it's quota-limited, and is actually handled in this part of my PR: https://github.com/PostHog/posthog-js-lite/pull/403/files#diff-e8ba078a1fb3324f02789b3d54af7bd54e24168943fcb1ffa6de45486fa6d760R465-R471
2. `/local_eval`, which will fail in a new way, but, like I mentioned earlier, is only scoped to flags, already fails with 400s, and is already handled top-level by all SDKs that leverage it.

Does this make sense?  Any additional feedback?

marandaneto: > The change we're talking about is a change to the /local_evaluation endpoint

Ahaaaa, I missed this point, sorry about the confusion, all good then.

posthog-core/src/index.ts
```diff
@@ -1225,7 +1249,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
     return this._decideResponsePromise
   }
 
-  private setKnownFeatureFlags(featureFlags: PostHogDecideResponse['featureFlags']): void {
+  private setKnownFeatureFlags(featureFlags: PostHogDecideResponse['featureFlags'] | null): void {
```
marandaneto, line 1252: since this and `setKnownFeatureFlagPayloads` accept `null` now, we can change the other places from:
```
this.setKnownFeatureFlags({})
this.setKnownFeatureFlagPayloads({})
```

to null values as well.

marandaneto (approved):
left a [comment](https://github.com/PostHog/posthog-js-lite/pull/403/files#r1966803342) but otherwise LGTM

marandaneto:
before merging, lets figure this out https://github.com/PostHog/posthog-android/pull/228#discussion_r1966811362

posthog-node/src/feature-flags.ts
marandaneto: pushed a fix [`eb8a5c7` (#403)](https://github.com/PostHog/posthog-js-lite/pull/403/commits/eb8a5c77ef282cc2b86b480eb24fac1e3d4d7744) to address disabling recording if no flags

posthog-core/src/index.ts
dmarticus: I did this I believe; I grepped those methods and I am using `null` in the only place they're called.

PR #405, opened 2025-02-22
Add KeyboardAvoidingView to SurveyModal

ian-craig:
Pulling over a couple of bug fixes from my fork:

- When focusing a text field in a popover/modal survey the keyboard would cover the modal so you couldn't see what you were typing
- The `useFeedbackSurvey` hook was returning surveys which were not launched/enabled in PostHog

#110

### DIFF
--- a/posthog-react-native/src/surveys/PostHogSurveyProvider.tsx
+++ b/posthog-react-native/src/surveys/PostHogSurveyProvider.tsx
@@ -32,7 +32,11 @@ const FeedbackSurveyContext = React.createContext<
 export const useFeedbackSurvey = (selector: string): FeedbackSurveyHook | undefined => {
   const context = React.useContext(FeedbackSurveyContext)
   const survey = context?.surveys.find(
-    (survey: Survey) => survey.type === SurveyType.Widget && survey.appearance?.widgetSelector === selector
+    (survey: Survey) =>
+      survey.type === SurveyType.Widget &&
+      survey.appearance?.widgetSelector === selector &&
+      survey.start_date !== null &&
+      survey.end_date === null
   )
   if (!context || !survey) {
     return undefined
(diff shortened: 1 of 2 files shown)


posthog-react-native/src/surveys/PostHogSurveyProvider.tsx
```diff
@@ -32,7 +32,11 @@ const FeedbackSurveyContext = React.createContext<
 export const useFeedbackSurvey = (selector: string): FeedbackSurveyHook | undefined => {
   const context = React.useContext(FeedbackSurveyContext)
   const survey = context?.surveys.find(
-    (survey: Survey) => survey.type === SurveyType.Widget && survey.appearance?.widgetSelector === selector
+    (survey: Survey) =>
+      survey.type === SurveyType.Widget &&
+      survey.appearance?.widgetSelector === selector &&
+      survey.start_date !== null &&
+      survey.end_date === null
```
marandaneto, line 39: I think we should call `getActiveMatchingSurveys` instead, passing `context?.surveys`.
Since the survey should only be shown in certain conditions, and this is a subset of it but, everything should apply right, even if you call `useFeedbackSurvey` yourself.

ian-craig: Yeah, I guess that makes sense... 🤔 

For context, my scenario for using this hook is that I have UI in our app where a customer can explicitly open a survey. e.g.

- A "Submit feedback" button in the menu which opens a generic feedback survey. If survey doesn't exist, fall back to emailing us
- A "Request another integration" button which appears at the end of a list of integrations. If survey doesn't exist, just don't show the button.

Because of the fallbacks I can't just use event triggers and a noop button, and I figure some devs may want a custom UI that just submits data back to posthog. I hadn't run into the need for all the conditions `getActiveMatchingSurveys` handles, but thinking about it now I don't think it'd hurt and may be useful for some people.

sepperousseau:
@marandaneto Is the KeyboardAvoidingView still on the radar?

Current implementation still has the problem that the keyboard goes over the text field in the modal.
The user can't see the input, and when return is pressed the input is submitted and closed.

marandaneto:
> @marandaneto Is the KeyboardAvoidingView still on the radar?
> 
> Current implementation still has the problem that the keyboard goes over the text field in the modal. The user can't see the input, and when return is pressed the input is submitted and closed.

sorry I don't get notified of messages on closed PRs.
similar issue here https://github.com/PostHog/posthog/issues/30858
taking a look at it

PR #408, opened 2025-02-24
feat(flags): Add guardrails and error backoff to local evaluation poller

havenbarnes:
### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -3,6 +3,7 @@ import { FeatureFlagCondition, FlagProperty, PostHogFeatureFlag, PropertyGroup }
 import { JsonType, PostHogFetchOptions, PostHogFetchResponse } from 'posthog-core/src'
 import { safeSetTimeout } from 'posthog-core/src/utils'
 import fetch from './fetch'
+import { SIXTY_SECONDS } from './posthog-node'
 
 // eslint-disable-next-line
 const LONG_SCALE = 0xfffffffffffffff
@@ -57,6 +58,8 @@ class FeatureFlagsPoller {
   debugMode: boolean = false
   onError?: (error: Error) => void
   customHeaders?: { [key: string]: string }
+  lastRequestWasAuthenticationError: boolean = false
+  authenticationErrorCount: number = 0
 
   constructor({
     pollingInterval,
@@ -78,7 +81,6 @@ class FeatureFlagsPoller {
     this.projectApiKey = projectApiKey
     this.host = host
     this.poller = undefined
-    // NOTE: as any is required here as the AbortSignal typing is slightly misaligned but works just fine
     this.fetch = options.fetch || fetch
     this.onError = options.onError
     this.customHeaders = customHeaders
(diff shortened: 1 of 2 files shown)


posthog-node/src/feature-flags.ts
```diff
@@ -375,19 +377,44 @@ class FeatureFlagsPoller {
...
+   * until a successful request is made, up to a maximum of 60 seconds.
+   * 
+   * @returns The polling interval to use for the next request.
+   */
+  private getPollingInterval(): number {
+    if (!this.lastRequestWasAuthenticationError) {
+      return this.pollingInterval
+    }
+
+    return Math.min(SIXTY_SECONDS, this.pollingInterval * 2 ** this.authenticationErrorCount)
```
havenbarnes, line 391: @MichaelKutsch-ph do ya think we should be more aggressive here, allowing it to bump higher than 60s?

MichaelKutsch-ph: Let's start like this and see how much that helps.

posthog-node/src/feature-flags.ts
```diff
@@ -375,19 +377,44 @@ class FeatureFlagsPoller {
...
+    this.poller = setTimeout(() => this._loadFeatureFlags(), this.getPollingInterval())
 
     try {
       const res = await this._requestFeatureFlagDefinitions()
 
       if (res && res.status === 401) {
+        this.lastRequestWasAuthenticationError = true
+        this.authenticationErrorCount += 1
+        throw new ClientError(
+          `Your project key or personal API key is invalid. Setting next polling interval to ${this.getPollingInterval()}ms. More information: https://posthog.com/docs/api/overview`
```
dmarticus, line 409: worth being even more specific and linking to the rate-limiting part of this page? 
```suggestion
          `Your project key or personal API key is invalid. Setting next polling interval to ${this.getPollingInterval()}ms. More information: https://posthog.com/docs/api#rate-limiting`
```

posthog-node/src/posthog-node.ts
```diff
@@ -47,12 +51,17 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
 
     this.options = options
 
+    this.options.featureFlagsPollingInterval = typeof options.featureFlagsPollingInterval === 'number'
+    ? Math.max(options.featureFlagsPollingInterval, MINIMUM_POLLING_INTERVAL) 
+    : THIRTY_SECONDS
+
     if (options.personalApiKey) {
+      if (options.personalApiKey.includes('phc_')) {
```
marandaneto, line 59: is that the case for API keys generated lets say years ago? just thinking of back-compatibility here

marandaneto (approved):
left a comment [here](https://github.com/PostHog/posthog-js-lite/pull/408/files#r1970996805) otherwise LGTM

posthog-node/src/posthog-node.ts
havenbarnes: Good point, I'll check metabase for the oldest ones

havenbarnes: Ok, `phc_` prefix was added at the same time as `phx_` prefix (see PR [here](https://github.com/PostHog/posthog/pull/4832/files#diff-4a33cc1373733c603cd71717b6f97addbf231f1e3a202422922ce860384819bfR113)). Before that it was just random alphanumeric content. This means that `phc_` was not always there for project keys, however personal API keys never contained `phc_` either, so this should be a safe check

haacked:
I wonder if we should do this on all the clients.

havenbarnes:
> I wonder if we should do this on all the clients.

Yes, I'm planning on adding it to python next!

PR #414, opened 2025-03-04
feat(flags): Allow feature flags to be evaluated individually

havenbarnes:
## Problem

For customers with high volume / high complexity of flags, fetching all flags when calling `isFeatureEnabled` or `getFeatureFlagPayload` is quite heavy handed. This PR forces any singular-flag lookup to ask only for it from `/decide` if decide is to be used. There should be no changes to local evaluation

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- feat(flags): Allow feature flags to be evaluated individually


dmarticus (approved):
let's get this fixed 🤞

PR #421, opened 2025-03-10
fix: update the rate-limiting logic to ensure we don't repeatedly fetch

frankh:
## Problem

We weren't handling backing off on 429s, so we kept hammering the `fetch` if a request was rate-limited.

## Changes

Added an error handler that wraps 429 errors in exponential backoffs (like we do with 401 and 403 errors)

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Fix: prevent fetch floods when quota limited.

### DIFF
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -418,7 +418,7 @@ class FeatureFlagsPoller {
         )
       }
 
-      if (res && res.status === 402) {
+      if (res && (res.status === 402 || res.status === 429)) {
         // Quota limited - clear all flags
         console.warn(
           '[FEATURE FLAGS] Feature flags quota limit exceeded - unsetting all local flags. Learn more about billing limits at https://posthog.com/docs/billing/limits-alerts'
@@ -427,7 +427,9 @@ class FeatureFlagsPoller {
         this.featureFlagsByKey = {}
         this.groupTypeMapping = {}
         this.cohorts = {}
-        this.loadedSuccessfullyOnce = false
+        // set loadedSuccessfullyOnce on 429s, otherwise every call to getFeatureFlag
+        // will try to fetch again.
+        this.loadedSuccessfullyOnce = true
         return
       }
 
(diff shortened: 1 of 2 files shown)


posthog-node/src/feature-flags.ts
```diff
@@ -427,10 +419,26 @@ class FeatureFlagsPoller {
         this.featureFlagsByKey = {}
         this.groupTypeMapping = {}
         this.cohorts = {}
-        this.loadedSuccessfullyOnce = false
+        this.loadedSuccessfullyOnce = true
         return
       }
 
+      if (res && res.status === 403) {
```
marandaneto, line 426: maybe just return early if `res` is undefined/null? we're checking for every condition

rafaeelaudibert: +1

rafaeelaudibert (approved):
LGTM, but I agree we could simplify the `res && res.status === something` checks. That can be a follow-up

posthog-node/src/feature-flags.ts
```diff
@@ -427,10 +419,26 @@ class FeatureFlagsPoller {
         this.featureFlagsByKey = {}
         this.groupTypeMapping = {}
         this.cohorts = {}
-        this.loadedSuccessfullyOnce = false
+        this.loadedSuccessfullyOnce = true
```
haacked, line 422: This line here is the main fix, correct? Might be worth a comment on why this is set to `true` here even though we're in a `402` status. I understand now, but at first, I thought this was a mistake.

dmarticus: actually, this line was not the fix that we needed, and I've since removed it.  There was some confusion about the original cause of this error – originally we thought it was from quota limited responses (402s), but turns out it was actually that we were repeatedly fetching _rate_-limited responses (429s).

So, the changes I made to that end were:
- wrap the rate limiting errors in the same exponential backoff as 401 and 403s
- continue to do the same thing with 402s, which is clear the flags but _don't_ return the error.

See my latest diff for more context on all of this

marandaneto (approved):

PR #423, opened 2025-03-12
fix: ensure we don't send multiple flag requests in parallel

frankh:
## Problem

currently if we call e.g. `getFeatureFlags()` 100 times straight away after initialisation we will send 100 fetch requests before the first one returns

this is even worse if e.g. we are erroring so never setting `loadedSuccessfullyOnce`

## Changes

add logic to check if we have an inflight request already and block on the existing one instead of starting a new one

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Fix: don't send multiple feature flag fetch requests in parallel.

### DIFF
--- a/posthog-node/src/feature-flags.ts
+++ b/posthog-node/src/feature-flags.ts
@@ -81,6 +81,7 @@ class FeatureFlagsPoller {
     this.projectApiKey = projectApiKey
     this.host = host
     this.poller = undefined
+    this.inflightRequest = undefined
     this.fetch = options.fetch || fetch
     this.onError = options.onError
     this.customHeaders = customHeaders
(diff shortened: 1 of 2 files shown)


havenbarnes (approved):
Nice, makes sense and posthog-js does this already

marandaneto:
@PostHog/team-feature-flags should this be finished/merged/released?

posthog-node/src/feature-flags.ts
```diff
@@ -396,11 +397,19 @@ class FeatureFlagsPoller {
       clearTimeout(this.poller)
       this.poller = undefined
     }
-
     this.poller = setTimeout(() => this._loadFeatureFlags(), this.getPollingInterval())
```
rafaeelaudibert, line 400: We might need to tweak this too, right? We could probably check for the `inflightRequest` **above** this

marandaneto:
moving to draft until we hear from @PostHog/team-feature-flags

PR #428, opened 2025-03-25
chore(flags): emit an error message whenever `/decide` fails to calculate flags

dmarticus:
## Problem

Customer asked for this, seems reasonable to do it: https://posthog.slack.com/archives/C0386ASS3TN/p1742599095502899?thread_ts=1740762972.509389&cid=C0386ASS3TN

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -522,7 +522,7 @@ export abstract class PostHogCoreStateless {
     }
     const decideResponse = await this.getDecide(distinctId, groups, personProperties, groupProperties, extraPayload)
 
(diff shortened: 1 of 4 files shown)


haacked:
Nice! We should do this across the board!

haacked (approved):
LGTM! Just need to accept the greptile suggestion.

dmarticus:
> Nice! We should do this across the board!

Yeah, probably.  Just doing this to make a customer happy but ultimately we should do this everywhere.

PR #430, opened 2025-03-25
fix: langchain name + usage

k11kirky:
## Problem

langchain is not using runName so $ai_span_name is incorrectly set

## Changes

Fix langchain name

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-ai
- [ ] posthog-react-native

### DIFF
--- a/posthog-ai/src/langchain/callbacks.ts
+++ b/posthog-ai/src/langchain/callbacks.ts
@@ -586,8 +586,6 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
     let llmUsage: [number, number, Record<string, any>] = [0, 0, {}]
     const llmUsageKeys = ['token_usage', 'usage', 'tokenUsage']
 
-    console.log('response1', response)
-
     if (response.llmOutput != null) {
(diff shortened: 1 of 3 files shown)


posthog-ai/src/langchain/callbacks.ts
```diff
@@ -520,7 +530,7 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
     return messageDict
   }
 
-  private _parseUsageModel(usage: any): [number, number] {
+  private _parseUsageModel(usage: any): [number, number, Record<string, any>] {
```
havenbarnes, line 533: nit: could we use `Record<string, number>` throughout instead?

havenbarnes (approved):

posthog-ai/src/langchain/callbacks.ts
k11kirky: hmmm, i am not confident it is always a number, but good thinking. will look into

PR #431, opened 2025-03-25
fix: Fallback to Web API Crypto when node:crypto not available

### LINKED ISSUE #344: UnhandledSchemeError `node:crypto`
## Bug description

Error deploying to Vercel on latest `4.4.0` release. It appears `node:crypto` was introduced in `4.4.0` (https://github.com/PostHog/posthog-js-lite/pull/329). I have locked my `posthog-node` version to `4.3.0` in the meantime as a workaround.

```
web:build: Failed to compile.
web:build: 
web:build: node:crypto
web:build: Module build failed: UnhandledSchemeError: Reading from "node:crypto" is not handled by plugins (Unhandled scheme).
web:build: Webpack supports "data:" and "file:" URIs by default.
web:build: You may need an additional plugin to handle "node:" URIs.
web:build: node:crypto
web:build: ../../node_modules/.pnpm/posthog-node@4.4.0/node_modules/posthog-node/lib/index.esm.js
web:build: ../../packages/plugins/analytics/posthog/src/index.ts
web:build: ../../packages/analytics/src/index.ts
web:build: ./components/analytics-provider.tsx
web:build: ./components/root-providers.tsx
web:build: 
web:build: 
web:build: > Build failed because of webpack errors
```

## How to reproduce

1. Install latest 4.4.0 `posthog-node` package
2. Deploy to Vercel

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

haacked:
## Problem

`node:crypto` doesn't work on edge runtimes such as Vercel and Cloudflare Pages.

Fixes #344 

## Changes

- Refactored the existing `_hash` method to rely on a new `hashSHA1` method.
- The `hashSHA1` method attempts to first use `node:crypto` by dynamically importing it
- If it doesn't exist, then it falls back to use Web Crypto API

Note that the code is a little confusing because both `node:crypto` and Web Crypto API have the same name, but are imported differently:

* `node:crypto`: `await import('crypto')`
* Web Crypto API: `const { webcrypto } = await import('crypto')`

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Fix: Issue with node:crypto running on edge runtimes


posthog-node/src/feature-flags.ts
```diff
@@ -180,34 +180,36 @@ class FeatureFlagsPoller {
...
-        const matchPayload = await this.computeFeatureFlagPayloadLocally(flag.key, matchValue)
-        if (matchPayload) {
-          payloads[flag.key] = matchPayload
-        }
-      } catch (e) {
-        if (e instanceof InconclusiveMatchError) {
-          // do nothing
-        } else if (e instanceof Error) {
-          this.onError?.(new Error(`Error computing flag locally: ${flag.key}: ${e}`))
+    await Promise.all(
```
rafaeelaudibert, line 183: Slight change in behavior, but I believe the previous was actually broken, we could return `fallbackToDecide` as false because we might not have run the promises before returning.

posthog-node/src/crypto.ts
```diff
@@ -0,0 +1,21 @@
+/// <reference lib="dom" />
+
+import { getNodeCrypto, getWebCrypto } from './crypto-helpers'
+export async function hashSHA1(text: string): Promise<string> {
```
rafaeelaudibert, line 4: Any chance we could cache the importing/crypto detection? This function runs pretty often. Maybe we could have a top-level await and abstract this away entirely?

haacked: Good idea. I'll try that.

posthog-node/src/feature-flags.ts
haacked: Yeah, I think it was incorrect. Not sure how it ever worked other than by happenstance.

posthog-node/src/posthog-node.ts
```diff
@@ -173,7 +173,9 @@ export class PostHog extends PostHogCoreStateless implements PostHogNodeV1 {
             additionalProperties[`$feature/${feature}`] = variant
           }
         }
-        const activeFlags = Object.keys(flags || {}).filter((flag) => flags?.[flag] !== false)
+        const activeFlags = Object.keys(flags || {})
+          .filter((flag) => flags?.[flag] !== false)
+          .sort()
```
haacked, line 178: This makes the tests pass and not be flaky, but I think it's actually the right thing to do. Makes it easy to compare different calls in the activity log.

dmarticus (approved):
tested on Vercel and it worked for me!  Nice work!

Issue #395, opened 2025-02-15
posthog-node@4.5 breaks build for Cloudflare


ccremer:
## Bug description

```
> Using @sveltejs/adapter-cloudflare
✘ [ERROR] Could not resolve "node:fs"

    node_modules/posthog-node/lib/index.esm.js:2:33:
      2 │ import { createReadStream } from 'node:fs';
        ╵                                  ~~~~~~~~~

  Cannot use "node:fs" when deploying to Cloudflare.


✘ [ERROR] Could not resolve "node:readline"

    node_modules/posthog-node/lib/index.esm.js:3:32:
      3 │ import { createInterface } from 'node:readline';
        ╵                                 ~~~~~~~~~~~~~~~

  Cannot use "node:readline" when deploying to Cloudflare.
```

## How to reproduce

1. Have a SvelteKit project
2. Use adapter-cloudflare and posthog-node
3. Import posthog-node somewhere in code
4. `npm run build`

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

## Additional context

Version 4.4.1 worked just fine.

marandaneto:
@PostHog/team-error-tracking

oliverb123:
Thanks for the report! [It looks like cloudflare workers aren't a nodejs compatible runtime by default, and you'll need to turn on the nodejs_compat flag](https://developers.cloudflare.com/workers/runtime-apis/nodejs/#nodejs-compatibility) in order to use npm packages designed for use with the nodejs runtime. [Based on cloudflares compatibility matrix, with this flag enabled, the imports listed should be supported.](https://workers-nodejs-compat-matrix.pages.dev)

ccremer:
I have this `nodejs_compat` flag already enabled, both in wrangler.toml, and in the dashboard.
```toml
compatibility_date = "2024-11-11"
compatibility_flags = [ "nodejs_compat" ]
```

![Image](https://github.com/user-attachments/assets/055464b0-747b-4756-832c-4cf51960b4d0)

However, I've found out that with adapter-cloudflare@4.9.0 it doesn't build, but it does @5.0.3.
I'm guessing since [5.0.0](https://github.com/sveltejs/kit/pull/13132/files) node support is much better, but it might be relevant to mention the minimum versions in your docs as well.

Anyway, it works for me now. Thanks!

daibhin:
[...]

PR #442, opened 2025-04-08
fix: edge runtime support

daibhin:
## Problem

Solves https://github.com/PostHog/posthog-js-lite/issues/395#issuecomment-2783365204

## Changes

- Lazy load filesystem packages

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- fix: do not access `fs` or `readline` in when not available e.g. edge environments


posthog-node/src/extensions/error-tracking/context-lines.ts
daibhin, line 1: Easier viewed with whitespace changes disabled

PR #446, opened 2025-04-09
feat: add autocapture pageviews with history api

lricoy:
## Problem

This Bluesky thread gave us the idea: https://bsky.app/profile/danabra.mov/post/3lmflnn62ys2y to have a tiny sdk option for simple navigation event capture.

This internal Slack thread also has some comments, it is my first SDK PR: https://posthog.slack.com/archives/C03P7NL6RMW/p1744225093576939

I've bundled a simple repo to test this out: https://lricoy.github.io/test-posthog-history/

## Changes

Added an option to attach listeners to the browser's history api and autocapture pageviews.

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [x] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added `trackHistoryEvents` option to automatically capture navigation events in single-page applications using the History API.

### DIFF
--- a/posthog-core/src/patch.ts
+++ b/posthog-core/src/patch.ts
@@ -1,0 +1,46 @@
+// import { patch } from 'rrweb/typings/utils'
+// copied from: https://github.com/PostHog/posthog-js/blob/main/src/extensions/replay/rrweb-plugins/patch.ts
+// which was copied from https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/rrweb/src/utils.ts#L129
+// which was copied from https://github.com/getsentry/sentry-javascript/blob/b2109071975af8bf0316d3b5b38f519bdaf5dc15/packages/utils/src/object.ts
+import { isFunction } from './utils'
+
+export function patch(
+  source: { [key: string]: any },
+  name: string,
+  replacement: (...args: unknown[]) => unknown
+): () => void {
+  try {
+    if (!(name in source)) {
+      return () => {
+        //
+      }
+    }
+
+    const original = source[name] as () => unknown
+    const wrapped = replacement(original)
+
+    // Make sure it's a function first, as we need to attach an empty prototype for `defineProperties` to work
+    // otherwise it'll throw "TypeError: Object.defineProperties called on non-object"
+    if (isFunction(wrapped)) {
+      wrapped.prototype = wrapped.prototype || {}
+      Object.defineProperties(wrapped, {
+        __posthog_wrapped__: {
+          enumerable: false,
+          value: true,
+        },
+      })
+    }
+
+    source[name] = wrapped
+
+    return () => {
+      source[name] = original
+    }
+  } catch {
+    return () => {
+      //
+    }
+    // This can throw if multiple fill happens on a global object like XMLHttpRequest
+    // Fixes https://github.com/getsentry/sentry-javascript/issues/2043
+  }
+}
--- a/posthog-web/README.md
+++ b/posthog-web/README.md
@@ -22,7 +22,6 @@ import PostHog from 'posthog-js-lite'
 const posthog = new PostHog('my-api-key', {
   /* options, e.g. for self-hosted users */
   // host: "https://my-posthog.app.com"
-  // trackHistoryEvents: true // Enable tracking of History API navigation events as pageviews
 })
 
 // Capture generic events
@@ -78,11 +77,11 @@ posthog.optIn() // Will start tracking
 
 Single-page applications (SPAs) typically use the History API (`pushState`, `replaceState`) for navigation instead of full page loads. By default, PostHog only tracks the initial page load.
 
-To automatically track navigation events in SPAs, enable the `trackHistoryEvents` option:
+To automatically track navigation events in SPAs, enable the `captureHistoryEvents` option:
 
 ```ts
 const posthog = new PostHog('my-api-key', {
-  trackHistoryEvents: true
+  captureHistoryEvents: true
 })
 ```
 
--- a/posthog-web/README.md
+++ b/posthog-web/README.md
@@ -22,6 +22,7 @@ import PostHog from 'posthog-js-lite'
 const posthog = new PostHog('my-api-key', {
   /* options, e.g. for self-hosted users */
   // host: "https://my-posthog.app.com"
+  // trackHistoryEvents: true // Enable tracking of History API navigation events as pageviews
 })
 
 // Capture generic events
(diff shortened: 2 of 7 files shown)


rafaeelaudibert (commented):
I do love how easy this is for those using the History API, excellent job.

Main question: on our main `posthog-js` library we put a lot of effort into making sure we're tracking time spent in session and blahblahblah for SPAs, by sending some extra information in follow-up `$pageview` events. Are we doing the same in this lib? If not, we'll need to port that to this lib as well


---

Things that popped in my head

`trackHistoryEvents` is not a property that exists in our main `posthog-js` library, and I'm afraid I wouldn't want this library to have **more** features than our main lib.

I'm in favor of adding this feature to both as that might simplify some of our NextJS setup. That might be a slightly bigger undertaking, though, since you'll need to update a lot of docs :)

For dan in specific, he should be able to simply use the `lite` version and call the tracking automatically while this isn't merged - just like you did.

lricoy:
> Main question: on our main `posthog-js` library we put a lot of effort into making sure we're tracking time spent in session and blahblahblah for SPAs, by sending some extra information in follow-up `$pageview` events. Are we doing the same in this lib? If not, we'll need to port that to this lib as well

That makes total sense, thanks! If approved, I was planning on porting that to posthog-js as well, as it seems useful enough. I'll probably add some more work to handle `$pageleave` as well and help with those session time calculations. It seems there are some work around that for the RN version of this lib.

I will take a look at those extra features on posthog-js as well, thanks!

posthog-web/README.md
```diff
@@ -22,6 +22,7 @@ import PostHog from 'posthog-js-lite'
 const posthog = new PostHog('my-api-key', {
   /* options, e.g. for self-hosted users */
   // host: "https://my-posthog.app.com"
+  // trackHistoryEvents: true // Enable tracking of History API navigation events as pageviews
```
pauldambra, line 25: > trackHistoryEvents is not a property that exists in our main posthog-js library, and I'm afraid I wouldn't want this library to have more features than our main lib.

i'm ok with adding this here if we then also add it in posthog-js... it's a super cool improvement

(we should definitely play with it running agains posthog locally since that's a SPA to figure out if it does anything weird but that's only me being cautious :))

---

in the main lib we tend to call config options `captureX` so might be worth keeping that here and calling it `captureHistoryEvents` seems nit-picky but the cognitive overload of config grows over time :)

lricoy: Will apply the changes, I've bundled this version into a repo to give it a few shots as well: https://lricoy.github.io/test-posthog-history/

posthog-web/src/posthog-web.ts
```diff
@@ -84,4 +88,40 @@ export class PostHog extends PostHogCore {
...
+  private setupHistoryEventTracking(): void {
+    const window = this.getWindow()
+    if (!window) {
+      return
+    }
+
+    const originalPushState = window.history.pushState
+    const originalReplaceState = window.history.replaceState
+
+    window.history.pushState = (state, title, url) => {
```
pauldambra, line 101: in the main lib we copied a patch function from rrweb, which was in turn originally copied from sentry

(but there are hundreds out there)

https://github.com/PostHog/posthog-js/blob/main/src/extensions/replay/rrweb-plugins/patch.ts

the main thing to copy from that would be marking the thing as wrapped

https://github.com/PostHog/posthog-js/blob/main/src/extensions/replay/rrweb-plugins/patch.ts#L24-L32

again seems nit-picky but I _really_ struggled to fix bugs between Angular's Zone detection stuff and replay because it was hard to track what was wrapped, so it's nice for the future traveller to be able to easily see that something isn't the original when debugging something

also means we can check for that and not wrap the function a second time if this gets called more than once

posthog-web/src/posthog-web.ts
```diff
@@ -84,4 +88,40 @@ export class PostHog extends PostHogCore {
...
+
+  // Capture navigation as pageview with only navigation_type
+  // URL and pathname come from getCommonEventProperties()
+  private captureNavigationEvent(navigationType: 'pushState' | 'replaceState' | 'popstate'): void {
+    const window = this.getWindow()
+    if (!window) {
+      return
+    }
+
+    this.capture('$pageview', { navigation_type: navigationType })
```
pauldambra, line 125: to Rafa's point about there being lots of good stuff on page view in the main lib 
we should check if there's anything we can very easily add here

but otherwise it's also fine for us to say "yep, lite doesn't do that, you can use the full fat posthog to get that"

pauldambra:
this is super cool though @lricoy , love it!

posthog-web/src/posthog-web.ts
lricoy: Checking this now to see if we can easily grab something there.

posthog-web/README.md
pauldambra:
```suggestion
  // captureHistoryEvents: true // Enable tracking of History API navigation events as pageviews
```

posthog-web/README.md
```diff
@@ -72,4 +73,23 @@ posthog.onFeatureFlag('my-feature-flag', (value) => {
...
+
+## History API Navigation Tracking
+
+Single-page applications (SPAs) typically use the History API (`pushState`, `replaceState`) for navigation instead of full page loads. By default, PostHog only tracks the initial page load.
+
+To automatically track navigation events in SPAs, enable the `trackHistoryEvents` option:
+
+```ts
+const posthog = new PostHog('my-api-key', {
+  trackHistoryEvents: true
```
pauldambra, line 85:
```suggestion
  captureHistoryEvents: true
```

pauldambra (approved):
couple of comment edits
i didn't run/test this but I can if helpful, your example website was great for seeing it run though! 🙌 

very excited for the posthog-js follow-up and then all the documentation simplication 🚀

posthog-core/src/patch.ts
```diff
@@ -0,0 +1,46 @@
...
+export function patch(
+  source: { [key: string]: any },
+  name: string,
+  replacement: (...args: unknown[]) => unknown
+): () => void {
+  try {
+    if (!(name in source)) {
+      return () => {
+        //
+      }
```
pauldambra, line 16: from a bundle size perspective i wonder how many times we have `no-op` as a bare declaration
if it's more than just this two we might get a saving from having noOp as a global 🤔

pauldambra: don't block here though

posthog-web/src/posthog-web.ts
lricoy: I think the features @rafaeelaudibert mentioned are mostly handled by the [PageViewManager/ScrollManager](https://github.com/PostHog/posthog-js/blob/a3da1787a9fdf98fcff1952f85dbb714183a2bf3/src/page-view.ts) which does not seem crazy to have it here as well. 

The page view durations and increments don't even need the `ScrollManager` bits; they are calculated based on the last pageview ref that is kept around on the `PageViewManager`.

But there is also session-related stuff that is way way simpler here than what we have on both [SessionIdManager](https://github.com/PostHog/posthog-js/blob/2c3eb709b3a4b5c158339f31871a60ddc2e952c9/src/sessionid.ts#L22) 
and [SessionPropsManager](https://github.com/PostHog/posthog-js/blob/3a81887e4b297bc7c6412c67b686f837d41e6adb/src/session-props.ts#L46)... So I am not sure how much we should port without documenting or keeping track of the differences.

WDYT @robbie-c ?

robbie-c: Hmm.

We don't actually use the scroll information or time on page in web analytics - so I'd be tempted to leave it out here actually!

We haven't exactly defined what posthog-js-lite is for, but one version I would propose would be: "the smallest bundle size possible where everything in web analytics still works", wdyt?

robbie-c: (I'm pretty sure we don't meet that goal right now, FWIW, e.g. I don't think we handle $pageleave events in this library)

pauldambra: i'd say "the smallest bundle where basic analytics and flags work"

that way we're not tied to chasing a feature set for a subset of customers who want feature X, y, and Z _and_ a tiny bundle

this is the "i can live with almost nothing" bubndle

lricoy:
Merging this as it is a simpler change, while https://github.com/PostHog/posthog-js/pull/1886 still waits for a ✅

PR #447, opened 2025-04-09
feat: Add method to wait for feature flag definitions for local evaluation to be loaded.

### LINKED ISSUE #445: feat(flags): Add a way to determine if local flags are cached
## Is your feature request related to a problem?

When bringing up new servers, there's a small chance that the local evaluation endpoint is having an outage just as the servers are coming up with the `posthog-node` client. This will cause the clients to fallback to remote evaluation.

It would be handy to have a way to check if local feature flags have been cached so that can be reported to a load balancer.

## Describe the solution you'd like

A new method (naming can be debated here):

```js
function localEvaluationFlagsLoaded(): boolean {
    return this.featureFlags.length > 0
}

async function waitForLocalEvaluationReady(): Promise<boolean> {}
```

## Describe alternatives you've considered

Considered a callback or event when local flags are loaded, but that wouldn't initiate a load and cache of local flag definitions.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

## Additional context

https://posthoghelp.zendesk.com/agent/tickets/28510

#### _Thank you_ for your feature request – we love each and every one!

You're welcome!

haacked:
This PR provides a way for users to determine if feature flag definitions for local evaluation are loaded.

## Problem

Fixes #445

## Changes

This adds a new `waitForLocalEvaluationReady` method that waits up to 30 seconds (configurable) for local evaluation feature flags to be loaded.

* It returns `true` immediately if flags are already loaded.
* It returns `false` immediately if local evaluation is not enabled.

This new method waits for the new `localEvaluationFlagsLoaded` event which is emitted when we retrieve feature flag definitions from the local evaluation endpoint.

This PR also adds a synchronous `isLocalEvaluationReady()` method that checks the feature flag poller to see if any flags are loaded.

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added new `waitForLocalEvaluationReady()` method used to wait for local evaluation flags to be loaded.
- Added new `localEvaluationFlagsLoaded` event which is emitted every time local evaluation flags are loaded.
- Added new `isLocalEvaluationReady()` method to quickly check if feature flags are loaded.

### DIFF
--- a/examples/example-node/server.ts
+++ b/examples/example-node/server.ts
@@ -23,6 +23,12 @@ const posthog = new PostHog(PH_API_KEY, {
   },
 })
 
+console.log('LOCAL EVALUATION READY RIGHT AFTER CREATION: ', posthog.isLocalEvaluationReady())
+
+posthog.on('localEvaluationFlagsLoaded', (count) => {
+  console.log('LOCAL EVALUATION READY (localEvaluationFlagsLoaded) EVENT EMITTED: flags count: ', count)
+})
+
 posthog.debug()
 
 Sentry.init({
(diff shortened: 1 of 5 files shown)


dmarticus (approved):
No notes, this is great.  Love the tests.

Issue #448, opened 2025-04-14
`posthog-node` v4.10.2+ breaks our TypeScript build


simon-abbott:
## Bug description

Trying to build with `posthog-node` v4.10.2 or greater causes the error:

```
node_modules/posthog-node/lib/index.d.ts:32:1 - error TS1046: Top-level declarations in .d.ts files must start with either a 'declare' or 'export' modifier.

32 enum SurveyPosition {
   ~~~~
```

## How to reproduce

1. Install `posthog-node` v4.10.2 or greater
2. Install the latest version of `typescript` (I tested against v5.8.3)
3. Use `posthog-node` in a `.ts` file somewhere in the project
4. Try to build

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

marandaneto:
@simon-abbott 

thanks for reporting, our CI didn't fail though, strange.
I'll add a quick workaround and cut a version.

simon-abbott:
Yeah, I'm surprised nobody else ran into this issue either, especially since I was able to reproduce it with a clean project. Thanks for the quick triage!

simon-abbott:
@marandaneto Unfortunately that didn't seem to fix the issue. I put together a [StackBlitz repro](https://stackblitz.com/edit/stackblitz-starters-4mx6ztx4?file=index.ts) to hopefully help.

marandaneto:
> [@marandaneto](https://github.com/marandaneto) Unfortunately that didn't seem to fix the issue. I put together a [StackBlitz repro](https://stackblitz.com/edit/stackblitz-starters-4mx6ztx4?file=index.ts) to hopefully help.

i did create an empty app locally and it worked, well... (probably TS versioning?)
let's give another try with https://github.com/PostHog/posthog-js-lite/pull/451

lucasheriques:
@marandaneto i didn't have the time today, but will find some time to check soon if you can't figure out

marandaneto:
@simon-abbott https://github.com/PostHog/posthog-js-lite/releases/tag/posthog-node-v4.11.6 i think this one should have worked, let us know if not, sorry about that.

simon-abbott:
That one seems to have worked! Thank you!

PR #452, opened 2025-04-15
move survey export top-level declarations

marandaneto:
## Problem

Hopefully fix https://github.com/PostHog/posthog-js-lite/issues/448

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [X] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for X


posthog-core/src/types.ts
```diff
@@ -258,3 +256,231 @@ export type EvaluationReason = {
   condition_index: number | undefined
   description: string | undefined
 }
+
+// survey types
```
marandaneto, line 260: moved all types from survey-types to types to avoid importing/missing module issue

rollup.config.js
```diff
@@ -67,11 +65,6 @@ const configs = ['posthog-node', 'posthog-web', 'posthog-ai'].reduce((acc, x) =>
       output: [{ file: `./${x}/lib/index.d.ts`, format: 'es' }],
       plugins: [
         dts(),
-        alias({
```
marandaneto, line 70: not needed since its all under the types file now

PR #458, opened 2025-04-16
chore(flags): roll flags to everyone except the special snowflakes

dmarticus:
## Problem

Do the same thing for our `posthog-js-lite`, `posthog-node`, and `posthog-react-native` SDKs that I did with [this PR](https://github.com/PostHog/posthog-python/commit/a1583f66275ac3d2dea1f62b1428b3f028d1f948) – use the new `/flags` service to evaluate feature flags (instead of `/decide`)

## Changes

- exclude the top 10 (by event count) customers for each SDK
- roll out flags to everyone else
- use a homemade token hashing function for obfuscating the API tokens; it's temporary and will be removed on the full rollout; I just didn't want to take on a dependency on a crypto lib just for a change like this.

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [x] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [x] posthog-react-native

### Changelog notes

chore: roll out new feature flag evaluation backend to majority of customers


posthog-core/src/utils.ts
```diff
@@ -1,5 +1,90 @@
 import { FetchLike } from './types'
 
+// Rollout constants
+export const NEW_FLAGS_ROLLOUT_PERCENTAGE = 1
+// The fnv1a hashes of the tokens that are explicitly excluded from the rollout
+// see https://github.com/PostHog/posthog-js-lite/blob/main/posthog-core/src/utils.ts#L84
+// are hashed API tokens from our top 10 for each category supported by this SDK.
+export const NEW_FLAGS_EXCLUDED_HASHES = new Set([
```
dmarticus, line 8: ran this query for each type of SDK event to get top 10 event counts.
```
SELECT
    JSONExtractString(properties, 'organization_id') AS organization_id,
    sum(JSONExtractInt(properties, '$EVENT')) AS $EVENT
FROM events
WHERE event = 'organization usage report'
 and toDate(timestamp) >= today() - INTERVAL 30 DAY AND toDate(timestamp) < today()
  AND (
    JSONExtractInt(properties, '$EVENT') > 0
  )
group by organization_id
order by $EVENT desc
limit 10
```

posthog-core/src/utils.ts
```diff
@@ -80,3 +165,31 @@ export function getFetch(): FetchLike | undefined {
 export const isFunction = function (f: any): f is (...args: any[]) => any {
   return typeof f === 'function'
 }
+
+// FNV-1a hash function
+// https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+// I know, I know, I'm rolling my own hash function, but I didn't want to take on
```
haacked, line 171: Also, in this case, you're not actually rolling your own. You're just porting an implementation of a known function.

haacked (approved):
:shipit:

PR #459, opened 2025-04-17
chore: add posthog-edge package

daibhin:
## Problem

We are currently chasing our tails on https://github.com/PostHog/posthog-js-lite/issues/395 which is largely because we're trying to use Node.js specific APIs in edge runtimes. In practice we should have a separate package for Edge runtimes

## Changes

- Adds a new package
- Makes the "PostHog Node" client a more generic "backend" client in the core package (hoping this doesn't balloon package size)
- Pull out the Node specific parts so that they are optional

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for X


daibhin:
@marandaneto just tagging you as you probably have the most context on the React Native package.

I'm looking to add a backend client to the core package which is affecting the RN package size because it does no form of tree shaking right now. Is there a reason we're not using Rollup for the RN package? I'd love to have it use the same flow so that we get the benefits of tree shaking and can make the package smaller. Conscious there might be some blocker I don't know about

marandaneto:
> @marandaneto just tagging you as you probably have the most context on the React Native package.
> 
> I'm looking to add a backend client to the core package which is affecting the RN package size because it does no form of tree shaking right now. Is there a reason we're not using Rollup for the RN package? I'd love to have it use the same flow so that we get the benefits of tree shaking and can make the package smaller. Conscious there might be some blocker I don't know about

the CI setup/builds are past me so I've never changed it, but treeshaking in react native is tricky since its not fully supported by react native itself.
there are different ways eg with expo, metro, hermes/babel, not sure about rollup though, never did/tried myself.

daibhin:
@marandaneto awesome, thanks for the context. It would be great to align everything so let me spend a little time on it today, because I obviously don't want to [ballon the package size as is the current state](https://github.com/PostHog/posthog-js-lite/pull/459#issuecomment-2819454222). Happy to give it a try knowing it's not a rabbit hole someone has already explored!

PR #462, opened 2025-04-22
chore: react native rollup

daibhin:
## Problem

The React Native package does not use Rollup and thus balloons in size when we make changes like https://github.com/PostHog/posthog-js-lite/pull/459 which move more code into the core package

## Changes

- Switch to exporting CJS and ESM modules (rather than the basic JS files)
- Use rollup to make RN output files

TODO:
- [ ] Test that the changes still compile and run in the example apps

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [x] posthog-react-native

### Changelog notes

- Added support for X


daibhin:
@marandaneto awesome, you're three steps ahead of me! My next question was going to be "how do I test the examples"

posthog-core/src/utils.ts
```diff
@@ -114,7 +114,7 @@ export async function retriable<T>(fn: () => Promise<T>, props: RetriableOptions
   for (let i = 0; i < props.retryCount + 1; i++) {
     if (i > 0) {
       // don't wait when it's the last try
-      await new Promise((r) => setTimeout(r, props.retryDelay))
+      await new Promise<void>((r) => setTimeout(r, props.retryDelay))
```
daibhin, line 117: Need to add the explicit type here because Node expects a `Node.Duration` whereas React Native returns a `number`

posthog-core/src/utils.ts
```diff
@@ -158,7 +158,7 @@ export const isError = (x: unknown): x is Error => {
 }
 
 export function getFetch(): FetchLike | undefined {
-  return typeof fetch !== 'undefined' ? fetch : typeof global.fetch !== 'undefined' ? global.fetch : undefined
+  return typeof fetch !== 'undefined' ? fetch : typeof globalThis.fetch !== 'undefined' ? globalThis.fetch : undefined
```
daibhin, line 161: `globalThis` is the cross platform version of `global` (Node / web only)

daibhin:
Making some really good progress here. Managed to get the example working on the simulator (proven by the 3.15.0 version shown)
Screenshot 2025-04-22 at 21 45 31

Added up the removed files in https://github.com/PostHog/posthog-js-lite/pull/462#issuecomment-2822449953. Looks like the old bundle was 56.762 kB and we're down to +41.2 kB for each of the CommonJS and ESM files. We could not ship the ESM files given it's only necessary if React Native Web is being used but including them to be safe. They shouldn't be used in a customers bundled app.

Last thing to solve is the failing test :)

posthog-react-native/jest.config.js
```diff
@@ -11,5 +11,7 @@ module.exports = {
   coverageDirectory: 'coverage',
   testPathIgnorePatterns: ['<rootDir>/lib/', 'node_modules', 'examples'],
   fakeTimers: { enableGlobally: true },
-  transformIgnorePatterns: [],
+  transformIgnorePatterns: [
```
daibhin, line 14: Long standing jest bug: https://github.com/react-native-community/upgrade-support/issues/152

posthog-react-native/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Next
 
+# 3.15.0 – 2025-04-22
```
```suggestion
# 3.15.0 – 2025-04-23
```

posthog-react-native/babel.config.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  presets: ['module:metro-react-native-babel-preset', '@babel/env', '@babel/preset-typescript'],
+  presets: ['module:metro-react-native-babel-preset', '@babel/env'],
```
marandaneto, line 2: is this intentional? this is still being set in the rollup.config.js, though

daibhin: I don't think we need it as part of the Jest tests which is the only place this config is being used. As you said it is still included as a babel preset in the rollup config

marandaneto (approved):

PR #474, opened 2025-04-28
feat: truncate vercel text to avoid "too large" events

k11kirky:
## Problem

We are allowing very alrge payloads through which are leading to misleading and hard to debug errors with the usage of vercel ai sdk

## Changes

Update processing of event input and output to truncate very large events. 

NOTE: thjis may be reverted in the future as we can handle larger events

### Bump level

- [ ] Major
- [x] Minor
- [] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-ai
- [ ] posthog-react-native


posthog-ai/src/utils.ts
```diff
@@ -108,6 +108,17 @@ export const withPrivacyMode = (client: PostHog, privacyMode: boolean, input: an
   return (client as any).privacy_mode || privacyMode ? null : input
 }
 
+export const truncate = (str: string): string => {
+  // limit large outputs by truncating to 200kb (approx 200k bytes)
+  const MAX_OUTPUT_SIZE = 200000
```
marandaneto, line 113: I'd move this out of the function so there's no need to allocate/deallocate

posthog-ai/src/utils.ts
```diff
@@ -108,6 +108,17 @@ export const withPrivacyMode = (client: PostHog, privacyMode: boolean, input: an
...
+  const buffer = Buffer.from(str, 'utf8')
```
marandaneto, line 114: let's move `utf8` to a const since its being repeated quite often, DRY

posthog-ai/src/vercel/middleware.ts
```diff
@@ -49,14 +49,23 @@ const mapVercelParams = (params: any): Record<string, any> => {
 }
 
 const mapVercelPrompt = (prompt: LanguageModelV1Prompt): PostHogInput[] => {
-  return prompt.map((p) => {
+  // normalize single inputs into an array of messages
+  let promptsArray: any[]
+  if (typeof prompt === 'string') {
+    promptsArray = [{ role: 'user', content: prompt }]
+  } else if (!Array.isArray(prompt)) {
```
marandaneto, line 56: just curious why we need to do this since `LanguageModelV1Prompt` is declared as `Array<LanguageModelV1Message>`.

posthog-ai/src/vercel/middleware.ts
```diff
@@ -49,14 +49,23 @@ const mapVercelParams = (params: any): Record<string, any> => {
 }
 
 const mapVercelPrompt = (prompt: LanguageModelV1Prompt): PostHogInput[] => {
-  return prompt.map((p) => {
+  // normalize single inputs into an array of messages
+  let promptsArray: any[]
+  if (typeof prompt === 'string') {
```
marandaneto, line 54: it'd be cool to create a util for this eg `isString(obj)` since we repeat this often and it is very verbose

posthog-ai/src/vercel/middleware.ts
```diff
@@ -113,25 +122,36 @@ const mapVercelPrompt = (prompt: LanguageModelV1Prompt): PostHogInput[] => {
...
+            size: file.size,
+            type: file.type,
+          })),
+        }
+      : {}),
   }
-  // if text and no object or reasoning, return text
   if (output.text && !output.object && !output.reasoning) {
-    return [{ content: output.text, role: 'assistant' }]
+    return [{ content: truncate(output.text as string), role: 'assistant' }]
```
marandaneto, line 147: are we sure this is a string? otherwise it'll throw an exception

posthog-ai/src/vercel/middleware.ts
```diff
@@ -113,25 +122,36 @@ const mapVercelPrompt = (prompt: LanguageModelV1Prompt): PostHogInput[] => {
...
   }
-  return [{ content: JSON.stringify(output), role: 'assistant' }]
+  // otherwise stringify and truncate
+  const jsonOutput = JSON.stringify(output)
```
marandaneto, line 150: this might throw eg `TypeError`, should we wrap in a try catch?

marandaneto:
@k11kirky left a few comments; address them as you feel is sensible, not a blocker.
3 side comments though.
* Truncating JSON is always tricky because, after truncation, it becomes a broken JSON, leading to a poor user experience whenever someone tries to make sense of it.
* For truncating, you need to allocate everything in memory, which increases the memory footprint. Although the size of messages is unknown, this can lead to an Out of Memory (OOM) error at some point.
* Error handling, some of these code paths can throw exceptions, I think we should wrap in try-catch and log to the console (instead of hard fail).

marandaneto (approved):

posthog-ai/src/vercel/middleware.ts
k11kirky: from testing it is not always an array

PR #481, opened 2025-04-29
Add gzip compression using Compression Streams API

### LINKED ISSUE #254: Support `gzip` compression
## Is your feature request related to a problem?

When using `captureMode=form`, the compression is `lz64` (base64).
When using `captureMode=json`, there's no compression.

## Describe the solution you'd like

When using `captureMode=json`, we could either use `gzip` by default if possible, or offer another config to allow or not allow the gzip compression.
JS SDK has an option `disable_compression: boolean`.

## Related sub-libraries

- [X] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-react-native

raikasdev:
## Problem

Adds GZIP compression support for `json` capturemode

Closes #254

## Changes

Simple wrapper around CompressionStream and a function to check if the class is available

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [x] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for GZIP compression in posthog-js-lite using Web APIs

### DIFF
--- a/posthog-core/src/gzip.ts
+++ b/posthog-core/src/gzip.ts
@@ -1,5 +1,6 @@
 /**
  * Older browsers and some runtimes don't support this yet
+ * This API (as of 2025-05-07) is not available on React Native.
  */
 export function isGzipSupported(): boolean {
   return 'CompressionStream' in globalThis


posthog-core/src/index.ts
```diff
@@ -839,10 +840,11 @@ export abstract class PostHogCoreStateless {
 
     const payload = JSON.stringify(data)
 
+    const canGzip = this.captureMode === 'json' && isGzipSupported();
```
marandaneto, line 843: we can call `isGzipSupported` once when the SDK is started instead of on every single flush.
we should also offer a way of disabling `gzip` even if its possible.

posthog-core/src/index.ts
```diff
@@ -839,10 +840,11 @@ export abstract class PostHogCoreStateless {
...
     const url =
       this.captureMode === 'form'
         ? `${this.host}/e/?ip=1&_=${currentTimestamp()}&v=${this.getLibraryVersion()}`
-        : `${this.host}/batch/`
+        : `${this.host}/batch/${canGzip ? '?compression=gzip-js' : ''}`
```
marandaneto, line 847: I think we dont need this actually, just `"Content-Encoding", "gzip"` header

marandaneto, line 847: not sure if we need `compression=gzip-js` here, I will need to check other SDKs.
we use `"Content-Encoding", "gzip"` header on Android/iOS but its not Web.

marandaneto:
@raikasdev the remote config/decide API will return `supportedCompression`, we need to check if its acceptable on the server side as well.
you can check the `PostHogDecideResponse` object.
and how this is used on the JS SDK: https://github.com/PostHog/posthog-js/blob/db26fffea82197b74e5f1f3c0b240b00d0e363e8/src/types.ts#L1255-L1258

posthog-core/src/index.ts
raikasdev: Because we send it using compressed in the body we need to use it if I remember correctly.
Because content-encoding "gzip"  uses the native HTTP transport compression :smile:

marandaneto: I think I found a few examples with `/s/?compression=gzip-js` URL but not the `/batch` so I don't know if the batch needs/accepts that.

raikasdev: I migrated to Content-Encoding because it's a bit more cleaner approach and perhaps easier for some self-hosters as the reverse proxy likely handles the encoding before it hits PostHog.

`/batch/` does accept it, at least the OpenAPI documentation says so [here](https://github.com/PostHog/posthog/blob/df24f8d1156987ebd45e05877e77968845e8fe5c/openapi/capture.yaml#L144-L153)

marandaneto:
@raikasdev once you fix tests (maybe you have to forcefully disable compression in tests)/lint I can test it out.
some unit tests would be awesome as well regarding the supported compressions and cangzip

raikasdev:
Fixed tests requiring JSON bodies and added few tests for gzip compress using Node's builtin gzip (via `node:zlib`)

posthog-core/src/gzip.ts
```diff
@@ -0,0 +1,25 @@
...
+export function isGzipSupported(): boolean {
+  return 'CompressionStream' in globalThis
+}
+
+/**
+ * Gzip a string using Compression Streams API if it's available
+ */
+export async function gzipCompress(input: string): Promise<Blob> {
+  if (!isGzipSupported()) {
+    throw new Error('Compression Streams API is not available in this runtime')
```
marandaneto, line 13: we should not throw here and return nothing so we can fallback to not compressed instead of failing hard

marandaneto: or wrap the whole method in a try catch so its safe to fail/log and move on

posthog-core/src/index.ts
```diff
@@ -855,8 +860,12 @@ export abstract class PostHogCoreStateless {
...
         : {
             method: 'POST',
-            headers: { ...this.getCustomHeaders(), 'Content-Type': 'application/json' },
-            body: payload,
+            headers: {
+              ...this.getCustomHeaders(),
+              'Content-Type': 'application/json',
+              ...(useGzip ? { 'Content-Encoding': 'gzip' } : {}),
+            },
+            body: useGzip ? await gzipCompress(payload) : payload,
```
marandaneto, line 868: we should only set `'Content-Encoding': 'gzip'` if `gzipCompress(payload)` also returns valid data.
so ideally you compute the `body` first and depending on the result, you set the encoding.

posthog-core/src/types.ts
```diff
@@ -128,13 +130,23 @@ export type PostHogAutocaptureElement = {
   [key: string]: any
 } // Any key prefixed with `attr__` can be added
 
+export enum Compression {
+  GZipJS = 'gzip-js',
+  Base64 = 'base64',
```
marandaneto, line 135: we should add `gzip` as well since might be needed for Node and React Native

marandaneto:
https://github.com/PostHog/posthog-js-lite/blob/df41c7bfba830024cc80060cc8626731bcac1f41/posthog-core/src/index.ts#L854
on this line we should also check for `disableCompression` and `supportedCompression`

posthog-core/src/index.ts
```diff
@@ -144,6 +147,7 @@ export abstract class PostHogCoreStateless {
     // Init promise allows the derived class to block calls until it is ready
     this._initPromise = Promise.resolve()
     this._isInitialized = true
+    this.disableCompression = !isGzipSupported() || (options?.disableCompression ?? false)
```
marandaneto, line 150: `lz64` is always supported and its a compression, so most likely we should not check for `isGzipSupported` here

posthog-core/src/index.ts
```diff
@@ -1439,6 +1448,10 @@ export abstract class PostHogCore extends PostHogCoreStateless {
               this.reloadFeatureFlags()
             }
 
+            if (!response.supportedCompression?.includes(Compression.GZipJS)) {
```
marandaneto, line 1451: we should check for lz64 here too

posthog-core/src/types.ts
```diff
@@ -52,6 +52,8 @@ export type PostHogCoreOptions = {
   sessionExpirationTimeSeconds?: number
   /** Whether to post events to PostHog in JSON or compressed format. Defaults to 'json' */
   captureMode?: 'json' | 'form'
+  /** Whether to disable GZIP compression */
+  disableCompression?: boolean
```
marandaneto, line 56: either this flag should consider lz64 or we should rename to be specifically about gzip compression

posthog-core/src/types.ts
raikasdev: GZipJS and GZIP are aliases to each other

marandaneto: true that, just found this here https://github.com/PostHog/posthog/blob/bcc624b7150124354edb4cd1f603cf73ce1dbfc9/posthog/api/decide.py#L118-L119

posthog-core/src/index.ts
marandaneto: nvm, https://github.com/PostHog/posthog/blob/bcc624b7150124354edb4cd1f603cf73ce1dbfc9/posthog/api/decide.py#L118-L119
its hardcoded and the API does not return `lz64` here

raikasdev:
Test should be fixed, weird that it worked locally without the import 😄. I'll look at the comments later today

posthog-core/src/gzip.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Older browsers and some runtimes don't support this yet
+ */
+export function isGzipSupported(): boolean {
+  return 'CompressionStream' in globalThis
```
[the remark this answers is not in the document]
marandaneto: if it works, all good, i've not tested yet just assumed its not the way in Node, that's what Google told me :D

posthog-core/src/index.ts
marandaneto: nvm, `lz64` is deprecated and will be removed at some point

posthog-core/src/types.ts
marandaneto: nvm, `lz64` is deprecated and will be removed at some point

raikasdev:
@marandaneto If `lz64` is getting deprecated should the captureMode form also be using gzip if it is available in the runtime?

posthog-core/src/gzip.ts
raikasdev: yep this now fails silently and returns null

posthog-core/src/gzip.ts
```diff
@@ -0,0 +1,27 @@
...
+    const dataStream = new Blob([input], {
+      type: 'text/plain',
+    }).stream()
+
+    const compressedStream = dataStream.pipeThrough(new CompressionStream('gzip'))
+
+    // Using a Response to easily extract the readablestream value. Decoding into a string for fetch
+    return await new Response(compressedStream).blob()
+  } catch (error) {
+    if (isDebug) console.error('Failed to gzip compress data', error)
```
marandaneto, line 24: you can use `logMsgIfDebug` instead

marandaneto:
@raikasdev thanks
one last comment and fix the lint and GG

marandaneto (approved):

posthog-core/src/gzip.ts
marandaneto: true its a different file, lets keep this for now then.
regarding being cleaner, I agree but its much easier to refactor once things change since we use the same function across different packages, etc.

posthog-core/src/gzip.ts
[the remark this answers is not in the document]
marandaneto: @raikasdev just unresolving this because it does not work on react native.
it does on web and node though.
we could add a comment that this is web/node only for now, or would you prefer to make it work on react native as well using a lib or something?

marandaneto:
> Weird, locally prettier doesn't show anything 🤔

probably versioning, lets see if this helps [`272e099` (#481)](https://github.com/PostHog/posthog-js-lite/pull/481/commits/272e099e6b19fa18e68ba98793948f12deaacb72)

posthog-core/src/gzip.ts
```diff
@@ -0,0 +1,31 @@
...
+  return 'CompressionStream' in globalThis
+}
+
+/**
+ * Gzip a string using Compression Streams API if it's available
+ */
+export async function gzipCompress(input: string, isDebug = true): Promise<Blob | null> {
+  if (!isGzipSupported()) {
+    return null
+  }
```
marandaneto, line 14: not needed since we already do `this.disableCompression = !isGzipSupported() || (options?.disableCompression ?? false)`, and we only call `gzipCompress` if `disableCompression` is false

raikasdev: Ah, this is from the pre try-catch. I'll remove it

posthog-core/src/gzip.ts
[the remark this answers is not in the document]
marandaneto: @raikasdev ping so we can get this merged/released.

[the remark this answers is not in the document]
raikasdev: @marandaneto Ah sorry didn't see this. I'd say this should be marked browser/web only as this does use the standard API for compression streams and there's not much that can be done if the Hermes runtime doesn't support it.

I looked into a few React native GZIP implementations and they all seem to use native bindings and the few libraries I found had only a few thousand downloads so 🤷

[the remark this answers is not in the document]
marandaneto: yeah thats what i thought.
mind adding a comment on `disableCompression`?
Also setting `disableCompression: true` [here](https://github.com/PostHog/posthog-js-lite/blob/b8372284df5489457d95b16b0e4b12a91dd66c26/posthog-react-native/src/posthog-rn.ts#L81) so its disabled by default on RN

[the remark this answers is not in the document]
raikasdev: Is the default necessary? If Hermes later on adds support it would then start working automatically. And is commenting it that necessary anyways because in the comment it's pointed out some runtimes don't support this yet, "some runtimes" including React Native now.

CompressionStream API is part of [WinterTC's](https://wintertc.org/) Minimum Common API spec (WinterTC is an ECMA Technical Committee) so I wouldn't be suprised if native support is added in the future.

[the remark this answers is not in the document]
marandaneto: its mostly the runtime cost of checking if gzip exists which is fine to keep as is tbh.
just a comment telling this is currently not supported on RN is fine

clemens:
Hey folks,

I think this deserves an entry in the changelog, maybe even as a breaking change. Reasoning: I have a test where I'm using `nock` to mock out my requests to PostHog and making some assertions on the body to be sent – which I'd argue is a very common thing to do. However, with GZIP compression enabled by default, I'm now getting a compressed body.

While this is easy enough to work around (basically doing some kind of equivalent of `new PostHog({ disableCompression: process.env.NODE_ENV === "test" })`), it's still a bit of a WTF moment, since this isn't included in the changelog. (Unless I've missed it or there's an upgrade guide that I've missed.)

Just my 2 cents here – apart from this, the change is obviously great and very welcome. :)

marandaneto:
> Hey folks,
> 
> 
> 
> I think this deserves an entry in the changelog, maybe even as a breaking change. Reasoning: I have a test where I'm using `nock` to mock out my requests to PostHog and making some assertions on the body to be sent – which I'd argue is a very common thing to do. However, with GZIP compression enabled by default, I'm now getting a compressed body.
> 
> 
> 
> While this is easy enough to work around (basically doing some kind of equivalent of `new PostHog({ disableCompression: process.env.NODE_ENV === "test" })`), it's still a bit of a WTF moment, since this isn't included in the changelog. (Unless I've missed it or there's an upgrade guide that I've missed.)
> 
> 
> 
> Just my 2 cents here – apart from this, the change is obviously great and very welcome. :)

Thanks for the feedback, we forgot the changelog entry indeed.
@PostHog/team-client-libraries  mind adding for node/web? Latest major version changes (on pto/phone rn)

PR #479, opened 2025-04-29
chore: add express and edge packages

daibhin:
### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for X

PR #482, opened 2025-04-29
edge specific exports

daibhin:
## Problem

Towards https://github.com/PostHog/posthog-js-lite/issues/395
Edge runtimes aren't supported by our node package right now

## Changes

- Add new rollup steps to build separate runtimes
- Add exports for each of the supported runtimes
- Conditionally import stack parsing files that rely on Node specific APIs

Evolution of
https://github.com/PostHog/posthog-js-lite/pull/459
https://github.com/PostHog/posthog-js-lite/pull/479

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for X


posthog-ai/src/anthropic/index.ts
```diff
@@ -1,5 +1,5 @@
 import AnthropicOriginal from '@anthropic-ai/sdk'
-import { PostHog } from 'posthog-node'
+import { PostHog } from 'posthog-node/src/entrypoints/index.node'
```
daibhin, line 2: @hpouillot I think it's ok because we're not changing existing behaviour but we should look into making this import to the AI SDK runtime agnostic

daibhin: We will need a similar pattern for all the framework packages we need to create

hpouillot: Yes. I can see two options to do that, having real package dependencies or importing only the type

daibhin:
The bundle sizing is a little weird because we changed the structure of the outputs.


module | before | after
-- | -- | --
Node (CJS) | 38.1 kB | 38.3 kB
Node (ESM) | 37.8 kB | 38.1 kB
Edge (CJS) | - | 33.9 kB
Edge (ESM) | - | 33.7 kB

posthog-ai/src/anthropic/index.ts
pauldambra: is it worth commenting why this import changed?
looks like the kind of thing someone could "tidy up" without realising the impact?

posthog-ai/tests/openai.test.ts
```diff
@@ -1,7 +1,7 @@
-import { PostHog } from 'posthog-node'
+import { PostHog } from 'posthog-node/src/entrypoints/index.node'
```
pauldambra, line 1: oh i see it's all over... ah, maybe don't comment all over the place then 🤣

pauldambra (approved):
started reviewing, realised that so long as you've run the various environments, then you can have a trust stamp 🚢

daibhin:
@pauldambra

> started reviewing, realised that so long as you've run the various environments, then you can have a trust stamp 🚢

image

posthog-web/src/posthog-web.ts
```diff
@@ -1,15 +1,10 @@
...
-} from '../../posthog-core/src'
+import { PostHogCore, PostHogFetchOptions, PostHogFetchResponse, PostHogPersistedProperty } from 'posthog-core'
 import { getContext } from './context'
 import { PostHogStorage, getStorage } from './storage'
 import { version } from '../package.json'
 import { PostHogOptions } from './types'
-import { getFetch } from 'posthog-core/src/utils'
-import { patch } from '../../posthog-core/src/patch'
+import { getFetch } from 'posthog-core'
+import { patch } from 'posthog-core/src/patch'
```
daibhin, line 7: @hpouillot any reason you left this one as the direct import rather than `posthog-core`? Noticed you changed all the others in https://github.com/PostHog/posthog-js-lite/pull/482/commits/4de88beead83d05afed633ca59f1f6523bbe1b00

hpouillot: No specific reason, looks like I missed this one

rollup.config.js
```diff
@@ -35,48 +67,52 @@ const configs = ['posthog-node', 'posthog-web', 'posthog-ai'].reduce((acc, x) =>
...
+    external: external(`./posthog-node/package.json`),
+    plugins: plugins('posthog-node'),
+  })
+})
+
+// We only build types from node as all types should be the same
+configs.push({
+  input: `./posthog-node/src/entrypoints/index.node.ts`,
+  output: [{ file: `./posthog-node/lib/index.d.ts`, format: 'es' }],
+  external: external('./posthog-ai/package.json'),
```
daibhin, line 109: I'm guessing this is a typo @hpouillot?
```suggestion
  external: external('./posthog-node/package.json'),
```

hpouillot: Yes, it should `posthog-node` here

rollup.config.js
```diff
@@ -35,48 +67,52 @@ const configs = ['posthog-node', 'posthog-web', 'posthog-ai'].reduce((acc, x) =>
...
     },
     {
-      input: `./${x}/lib/${x}/index.d.ts`,
+      input: `./${x}/index.ts`,
       output: [{ file: `./${x}/lib/index.d.ts`, format: 'es' }],
-      plugins: [
-        dts(),
-      ],
+      external: external(`./${x}/package.json`),
+      plugins: [resolve({ extensions }), dts({ tsconfig: `./${x}/tsconfig.json` })],
```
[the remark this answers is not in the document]
hpouillot: It resolves (bundles) imports statements and as we use it with our external (all package dependencies), it will only bundle workspace packages.

rollup.config.js
```diff
@@ -35,48 +67,52 @@ const configs = ['posthog-node', 'posthog-web', 'posthog-ai'].reduce((acc, x) =>
...
-            '@babel/preset-react',
-          ],
-        }),
-      ],
+      external: external(`./${x}/package.json`),
+      plugins: plugins(x),
     },
     {
-      input: `./${x}/lib/${x}/index.d.ts`,
+      input: `./${x}/index.ts`,
```
daibhin, line 74: Commenting here as a mental reference for myself: We were relying on the declaration files produced in the previous step to produce a declaration output at the same path. This was essentially a no-op. Now that `declaration: false` is the default in our tsconfig we actually need to build the declaration file from the source (`<package>/index.ts`)

daibhin: Good catch. Done

hpouillot (approved):

posthog-web/src/patch.ts
daibhin, line 1: @hpouillot this file was only used in the web package and was increasing the RN bundle because we started exporting it from core. Moving it here instead

spelexander:
Awesome work. Glad to see this is in! 

However, I'm having an issue resolving the edge module by default in Cloudflare's workerd edge environment during builds with wrangler (still using lib/node by the looks of it). Would be nice if there's an obvious way to explicitly import the edge/node/other to avoid runtime detection problems.

daibhin:
@hpouillot do you know anything about specifying the runtime in Cloudflare environments (or more generally in the `package.json`)

spelexander:
Just trying out 4.18.0 and everthing is working as expected! Thanks

PR #484, opened 2025-04-30
fix: add immediate-mode

oliverb123:


posthog-core/src/index.ts
```diff
@@ -92,6 +92,7 @@ export abstract class PostHogCoreStateless {
   private removeDebugCallback?: () => void
   private disableGeoip: boolean
   private historicalMigration: boolean
+  private immediate: boolean
```
pauldambra, line 95: `sendImmediate`?

posthog-core/src/index.ts
```diff
@@ -720,33 +722,18 @@ export abstract class PostHogCoreStateless {
    *** QUEUEING AND FLUSHING
    ***/
   protected enqueue(type: string, _message: any, options?: PostHogCaptureOptions): void {
```
pauldambra, line 724: could allow a callback here
then convert that to a promise in some interfaces and it's not breaking for others?

(have not thought this through though!)

pauldambra: silly question though

if we set `maxBatchSize` to 1 is that (almost) the same as this?

marandaneto: yeah since it does not return a promise to be awaited, its the same as `maxBatchSize=1`

posthog-node/src/types.ts
```diff
@@ -79,6 +79,16 @@ export type PostHogNodeV1 = {
    */
   capture({ distinctId, event, properties, groups, sendFeatureFlags }: EventMessage): void
 
+  /**
```
marandaneto, line 82: should we add comments to it as experimental/might be removed in the future?

oliverb123: I don't think so? I can't think of a reason we would remove this, but if we do, it'll need to be a breaking change either way

oliverb123: real NODE_INTERNAL_DONT_USE_THIS_OR_YOU_WILL_BE_FIRED energy to put "experimental" in a comment imo

posthog-node/src/types.ts
```diff
@@ -79,6 +79,16 @@ export type PostHogNodeV1 = {
...
+   * @description Capture an event immediately. This is useful for events that should be captured immediately, such as a user logging in.
```
robbie-c, line 83: Nit: why is this in particular a good example?

robbie-c: What happens if a user mixes immediate and non-immediate? Presumably events can arrive out-of-order, which might have implications for person properties. IMO this is fine providing we tell people to only use one or the other, don't mix and match

oliverb123: It's not, removed

oliverb123: And the comment updated to tell people not to mix immediate and non-immediate calls.

robbie-c (approved):
LGTM - caveat that I haven't spent much time in this codebase before, but it seems like everything I'd expect is covered

PR #486, opened 2025-04-30
fix: Add a guard around calling shutdown multiple times

robbie-c:
## Problem

Calling shutdown multiple times can have an unexpected result. Additionally, we tell people to do this in our docs.

## Changes

Guard shutdown with a shared promise, similar to how we handle flush. Doesn't break anything if people are calling shutdown() per-request, but also gives them a debug log (and a comment) telling them not to do this

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [x] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added protection against calling shutdown in parallel, and update examples to be clearer on using flush() rather than shutdown() for serverless use-cases.

### DIFF
--- a/examples/example-gcp-functions/index.js
+++ b/examples/example-gcp-functions/index.js
@@ -1,37 +1,34 @@
-const functions = require('@google-cloud/functions-framework');
-const { PostHog } = require('posthog-node');
+const functions = require('@google-cloud/functions-framework')
+const { PostHog } = require('posthog-node')
 
 const posthog = new PostHog('phc_pQ70jJhZKHRvDIL5ruOErnPy6xiAiWCqlL4ayELj4X8', {
-    // works as well if you uncomment the following lines
-    // flushAt: 1,
-    // flushInterval: 0
-  })
+  // works as well if you uncomment the following lines
+  // flushAt: 1,
+  // flushInterval: 0
+})
 posthog.debug(true)
 
 async function sendEvent(id) {
-    // works as well if you uncomment the following line, and comment the global posthog declaration
-    // const posthog = new PostHog('phc_pQ70jJhZKHRvDIL5ruOErnPy6xiAiWCqlL4ayELj4X8')
-    // posthog.debug(true)
+  // works as well if you uncomment the following line, and comment the global posthog declaration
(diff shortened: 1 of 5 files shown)


examples/example-gcp-functions/index.js
```diff
@@ -1,37 +1,34 @@
-const functions = require('@google-cloud/functions-framework');
-const { PostHog } = require('posthog-node');
+const functions = require('@google-cloud/functions-framework')
```
robbie-c, line 1: most formatting changes, except for swapping shutdown() with flush()

posthog-core/test/posthog.shutdown.spec.ts
```diff
@@ -45,5 +45,45 @@ describe('PostHog Core', () => {
         })
       expect(mocks.fetch).toHaveBeenCalledTimes(1)
     })
+
+    it('return the same promise if called multiple times in parallel', async () => {
+      mocks.fetch.mockImplementation(async () => {
+        await new Promise((resolve) => setTimeout(resolve, 1000))
+        console.log('FETCH RETURNED')
```
rafaeelaudibert, line 52: This can go

rafaeelaudibert:
```suggestion
```

posthog-core/src/index.ts
```diff
@@ -962,6 +963,26 @@ export abstract class PostHogCoreStateless {
...
+
+  /**
+   *  Call shutdown() once before the node process exits, so ensure that all events have been sent and all promises
+   *  have resolved. Do not use this function if you intend to keep using this PostHog instance after calling it.
+   * @param shutdownTimeoutMs
+   */
+  async shutdown(shutdownTimeoutMs: number = 30000): Promise<void> {
+    if (this.shutdownPromise) {
+      this.logMsgIfDebug(() =>
+        console.error(
```
rafaeelaudibert, line 975: Is it really an error? More like a warning because we're gracefully handling it

rafaeelaudibert (approved):

marandaneto (approved):

PR #488, opened 2025-05-01
feat: Initial attempt at flush cleaning all of their data

rafaeelaudibert:
WIP

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -83,7 +83,7 @@ export async function logFlushError(err: any): Promise<void> {
     let text = ''
     try {
       text = await err.text
-    } catch {}
+    } catch { }
 
     console.error(`Error while flushing PostHog: message=${err.message}, response body=${text}`, err)
   } else {
@@ -256,7 +256,7 @@ export abstract class PostHogCoreStateless {
     const promiseUUID = uuidv7()
     this.pendingPromises[promiseUUID] = promise
     promise
-      .catch(() => {})
+      .catch(() => { })
       .finally(() => {
         delete this.pendingPromises[promiseUUID]
       })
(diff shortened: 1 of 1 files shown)


posthog-core/src/index.ts
```diff
@@ -912,57 +912,41 @@ export abstract class PostHogCoreStateless {
...
+    this.setPersistedProperty<PostHogQueueItem[]>(PostHogPersistedProperty.Queue, [])
 
-    const persistQueueChange = (): void => {
-      const refreshedQueue = this.getPersistedProperty<PostHogQueueItem[]>(PostHogPersistedProperty.Queue) || []
-      this.setPersistedProperty<PostHogQueueItem[]>(PostHogPersistedProperty.Queue, refreshedQueue.slice(items.length))
+    // Iterate over the queue and flush it in batches of maxBatchSize
+    const messages: PostHogQueueItem['message'][] = []
+    for (let i = 0; i < queue.length; i += this.maxBatchSize) {
+      const items = queue.slice(i, i + this.maxBatchSize)
+      messages.push(...(await this._flush(items)))
```
marandaneto, line 931: if it fails, it'll throw and lose all messages

posthog-core/src/index.ts
```diff
@@ -912,57 +912,41 @@ export abstract class PostHogCoreStateless {
...
-    const customUserAgent = this.getCustomUserAgent()
-    const headers: { [key: string]: string } = {}
-    if (customUserAgent && customUserAgent !== '') {
-      headers['User-Agent'] = customUserAgent
-    }
-    return headers
-  }
-
-  private async _flush(): Promise<any[]> {
+  async flush(): Promise<PostHogQueueItem['message'][]> {
```
marandaneto, line 915:
```suggestion
  async flush(): Promise<void> {
```

it's a breaking change anyways and we don't use the return obj anywhere

rafaeelaudibert:
@marandaneto this can be closed, Robbie solved this in a different way in #489

PR #489, opened 2025-05-01
fix: Improve flush behaviour

robbie-c:
## Problem

There's a long internal thread here: https://posthog.slack.com/archives/C087XQ7K9K7/p1745963996974469

We are not respecting the maximum body size restrictions on the capture endpoint, and `flush()` does not correctly flush all events.

## Changes

* Handle 413 errors, which can be returned if the request is too large. Handle this by reducing the batch size by half and retrying. If batch size was already 1, treat it as a regular error.
* Change the behaviour of `flush()`, to send >1 batch if necessary. Previously, it only sent 1 batch
* Ensure that `flush()` does not get stuck infinitely if there are a constant stream of new events. It is only guaranteed to flush events that occurred before `flush()` was called (and *may* include some afterwards, but is not guaranteed to) or until an error occurred.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [x] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Improved flush() behaviour, to ensure that it is more aware of batch size restrictions, and fix an infinite loop.

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -92,8 +92,12 @@ export async function logFlushError(err: any): Promise<void> {
   return Promise.resolve()
 }
 
-function isPostHogFetchError(err: any): boolean {
+function isPostHogFetchError(err: unknown): err is PostHogFetchHttpError | PostHogFetchNetworkError {
   return typeof err === 'object' && (err instanceof PostHogFetchHttpError || err instanceof PostHogFetchNetworkError)
+}
+
+function isPostHogFetchContentTooLargeError(err: unknown): err is PostHogFetchHttpError & { status: 413 } {
+  return typeof err === 'object' && err instanceof PostHogFetchHttpError && err.status === 413
 }
 
 enum QuotaLimitedFeature {
@@ -960,7 +964,7 @@ export abstract class PostHogCoreStateless {
     this.clearFlushTimer()
     await this._initPromise
 
-    const queue = this.getPersistedProperty<PostHogQueueItem[]>(PostHogPersistedProperty.Queue) || []
+    let queue = this.getPersistedProperty<PostHogQueueItem[]>(PostHogPersistedProperty.Queue) || []
 
     if (!queue.length) {
       return []
@@ -975,10 +979,9 @@ export abstract class PostHogCoreStateless {
 
       const persistQueueChange = (): void => {
         const refreshedQueue = this.getPersistedProperty<PostHogQueueItem[]>(PostHogPersistedProperty.Queue) || []
-        this.setPersistedProperty<PostHogQueueItem[]>(
-          PostHogPersistedProperty.Queue,
-          refreshedQueue.slice(batchItems.length)
-        )
+        const newQueue = refreshedQueue.slice(batchItems.length)
+        this.setPersistedProperty<PostHogQueueItem[]>(PostHogPersistedProperty.Queue, newQueue)
+        queue = newQueue
       }
 
       const data: Record<string, any> = {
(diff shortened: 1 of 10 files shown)


posthog-core/src/index.ts
```diff
@@ -923,13 +927,24 @@ export abstract class PostHogCoreStateless {
   }
 
   async flush(): Promise<any[]> {
-    if (!this.flushPromise) {
-      this.flushPromise = this._flush().finally(() => {
+    // Wait for the current flush operation to finish (regardless of success or failure), then try to flush again.
+    // Use allSettled instead of finally to be defensive around flush throwing errors immediately rather than rejecting.
+    const nextFlushPromise = Promise.allSettled([this.flushPromise]).then(() => {
```
rafaeelaudibert, line 932: What happens when `this.flushPromise` is null/undefined? Does `allSettled` work with that?

posthog-core/src/index.ts
```diff
@@ -949,69 +964,100 @@ export abstract class PostHogCoreStateless {
...
-      this._events.emit('error', err)
 
-      throw err
-    }
+      try {
+        await this.fetchWithRetry(url, fetchOptions, retryOptions)
+      } catch (err) {
+        if (isPostHogFetchContentTooLargeError(err) && batchMessages.length > 1) {
+          // if we get a 413 error, we want to reduce the batch size and try again
+          this.maxBatchSize = Math.max(1, Math.floor(batchMessages.length / 2))
```
rafaeelaudibert, line 1035: This changes the batch size forever, but this could have happened because of a single big message. What if we had an internal ephemeral batch size instead?

posthog-core/test/posthog.flush.spec.ts
```diff
@@ -1,5 +1,7 @@
 import { createTestClient, PostHogCoreTestClient, PostHogCoreTestClientMocks } from './test-utils/PostHogCoreTestClient'
-import { waitForPromises } from './test-utils/test-utils'
+import { delay, waitForPromises } from './test-utils/test-utils'
+import { PostHogPersistedProperty } from '../src'
+import useRealTimers = jest.useRealTimers
```
rafaeelaudibert, line 4: Is this valid TS?

posthog-node/test/posthog-node.spec.ts
```diff
@@ -381,33 +381,29 @@ describe('PostHog Node.js', () => {
...
-      expect(1).toEqual(logSpy.mock.calls.filter((call) => call[1].includes('flush')).length)
-
-      logSpy.mockClear()
-      expect(logSpy).toHaveBeenCalledTimes(0)
-
-      console.warn('YOO!!')
+      ph.capture({ event: 'test-event-1', distinctId: '123' })
+
+      // start flushing, but don't wait for promise to resolve before resuming events
+      void ph.flush()
```
rafaeelaudibert, line 387: Could store this in a var and assert it was resolved by `shutdown`

posthog-core/src/index.ts
robbie-c: Yeah it does! I initially handled the null case separately but it turned out that I didn't need to so I just simplified the code

posthog-core/src/index.ts
robbie-c: I did wonder about that, but it would mean that we'd potentially end up needing to send each batch multiple times. It could "learn" over time (hey this sounds a bit like how TCP MTU works) but I wanted to keep it simple for now.

posthog-core/test/posthog.flush.spec.ts
robbie-c: Oh I think it is but I don't like it, I think pycharm probably  autogenerated it, I'll remove it

rafaeelaudibert (approved):
CI is failing with some untelated posthog-ai stuff, lgtm now! Excellent solution

posthog-node/test/posthog-node.spec.ts
robbie-c: Done - this felt pretty hacky but it's test code so at least if it breaks in a version update, it'll only affect us and not users

k11kirky (approved):

Issue #491, opened 2025-05-02
Import error when importing Posthog on latest release


jeff-an:
## Bug description

The release a couple of hours ago broke a lot of CI pipelines and our customers' CI pipelines. Seems related to the various bundling [changes](https://github.com/PostHog/posthog-js-lite/pull/482). The error we see is:

```
SyntaxError: Named export 'PostHog' not found. The requested module 'posthog-node' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'posthog-node';
const {try{if(!n)return n;if(typeof n=="string")return Me(n,3e4);if(typeof n=="object"){let{jsonString:t}=Li(JSON.stringify(n,(r,o)=>e?.includes(r)||typeof o=="string"&&o.length>3e4?"REDACTED":o),2e4);return JSON.parse(t)}return n}catch{return n}}import{cloneDeep: WG}from"lodash-es";var xR=/^(?!.*\S+\s+\S+).*$/,TR=/^[a-zA-Z0-9\s.,;:'"!?()\- \u00C0-\u00FF]+$/,ER=/style="([^"]*)/g,CR=/data-[\w-]+/g,RR=n=>{switch(n.type){":if(n
Node.js v21.7.3
Error: Process completed with exit code 1.
```

## How to reproduce

```
import { PostHog } from "posthog-node";
```

we use tsup to bundle posthog as part of a CLI. Here are relevant parts of our tsup config:

```
export default defineConfig([
  {
    entry: ["src/cli.ts"],
    format: ["esm"],
    outDir: "bin",
    clean: true,
    dts: false,
    external: externalPkgs,
    noExternal: internalPkgs,
    minify: true,
    define: {
      "process.env.POSTHOG_ANALYTICS_PROJECT_ID": JSON.stringify(
        process.env.POSTHOG_ANALYTICS_PROJECT_ID,
      ),
      "process.env.POSTHOG_ANALYTICS_WRITE_KEY": JSON.stringify(
        process.env.POSTHOG_ANALYTICS_WRITE_KEY,
      ),
      "process.env.POSTHOG_WEB_PROJECT_ID": JSON.stringify(
        process.env.POSTHOG_WEB_PROJECT_ID,
      ),
      "process.env.POSTHOG_WEB_WRITE_KEY": JSON.stringify(
        process.env.POSTHOG_WEB_WRITE_KEY,
      ),
[...]

PR #492, opened 2025-05-02
fix(posthog-node): give explicit file types

hpouillot:
## Problem

Related to : https://github.com/PostHog/posthog-js-lite/issues/491
- Module types are not detected correcly

## Changes

- Give filenames explicit types

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Fix tsup bundling

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -10,4 +10,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: preactjs/compressed-size-action@v2
         with:
-          pattern: 'posthog-{web,node,react-native,ai}/lib/**/*.js'
+          pattern: 'posthog-{web,node,react-native,ai}/lib/**/*.{cjs,mjs,js}'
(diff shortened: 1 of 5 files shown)


daibhin (approved):
Happy to approve if you're reasonably confident this fixes things.

Don't know if I fully understand how our changes changed the file names 🤔

You will need to update the changelog and bump the version if you want to release tonight

wardpeet (commented):
I was just going to do the same PR, this one works for both use cases!

Basically export maps do not work in non latest versions of node if you do not use the correct extensions. If you're not setting type module on your package you should make all esm files end with mjs.

https://nodejs.org/en/blog/release/v20.19.0#module-syntax-detection-is-now-enabled-by-default

wardpeet:
Here is a codesanbox that demonstrates the fix using patch-package. You can run node index.mjs and node index.js to see both working.
https://codesandbox.io/p/devbox/tjcjnn

if you remove patch-package you can see the bug

wardpeet:
@hpouillot you have a typo in your pr title

PR #493, opened 2025-05-05
chore: improve event types

marandaneto:
### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -18,6 +18,7 @@ import {
   FeatureFlagDetail,
   Survey,
   SurveyResponse,
+  PostHogGroupProperties,
 } from './types'
 import {
   createDecideResponseFromFlagsAndPayloads,
@@ -65,7 +66,7 @@ class PostHogFetchHttpError extends Error {
     return this.response.text()
   }
 
-  get json(): Promise<any> {
+  get json(): Promise<JsonType> {
     return this.response.json()
   }
 }
@@ -201,7 +202,7 @@ export abstract class PostHogCoreStateless {
     this._initPromise.then(() => fn())
   }
 
-  protected getCommonEventProperties(): any {
+  protected getCommonEventProperties(): PostHogEventProperties {
     return {
       $lib: this.getLibraryId(),
       $lib_version: this.getLibraryVersion(),
@@ -248,7 +249,11 @@ export abstract class PostHogCoreStateless {
     return this.disabled
   }
 
-  private buildPayload(payload: { distinct_id: string; event: string; properties?: PostHogEventProperties }): any {
+  private buildPayload(payload: {
+    distinct_id: string
+    event: string
+    properties?: PostHogEventProperties
+  }): PostHogEventProperties {
     return {
       distinct_id: payload.distinct_id,
       event: payload.event,
@@ -314,7 +319,7 @@ export abstract class PostHogCoreStateless {
   protected captureStateless(
     distinctId: string,
     event: string,
-    properties?: { [key: string]: any },
+    properties?: PostHogEventProperties,
     options?: PostHogCaptureOptions
   ): void {
     this.wrap(() => {
@@ -326,7 +331,7 @@ export abstract class PostHogCoreStateless {
   protected async captureStatelessImmediate(
     distinctId: string,
     event: string,
-    properties?: { [key: string]: any },
+    properties?: PostHogEventProperties,
     options?: PostHogCaptureOptions
   ): Promise<void> {
     const payload = this.buildPayload({ distinct_id: distinctId, event, properties })
@@ -336,7 +341,7 @@ export abstract class PostHogCoreStateless {
   protected aliasStateless(
     alias: string,
     distinctId: string,
-    properties?: { [key: string]: any },
+    properties?: PostHogEventProperties,
     options?: PostHogCaptureOptions
   ): void {
     this.wrap(() => {
@@ -357,7 +362,7 @@ export abstract class PostHogCoreStateless {
   protected async aliasStatelessImmediate(
     alias: string,
     distinctId: string,
-    properties?: { [key: string]: any },
+    properties?: PostHogEventProperties,
     options?: PostHogCaptureOptions
   ): Promise<void> {
     const payload = this.buildPayload({
@@ -779,7 +784,7 @@ export abstract class PostHogCoreStateless {
     this._props = val
   }
 
-  async register(properties: { [key: string]: any }): Promise<void> {
+  async register(properties: PostHogEventProperties): Promise<void> {
     this.wrap(() => {
       this.props = {
         ...this.props,
@@ -886,7 +891,7 @@ export abstract class PostHogCoreStateless {
     }
   }
 
-  private prepareMessage(type: string, _message: any, options?: PostHogCaptureOptions): any {
+  private prepareMessage(type: string, _message: any, options?: PostHogCaptureOptions): PostHogEventProperties {
     const message = {
       ..._message,
       type: type,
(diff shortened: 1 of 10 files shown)


posthog-core/src/index.ts
```diff
@@ -314,7 +319,7 @@ export abstract class PostHogCoreStateless {
   protected captureStateless(
     distinctId: string,
     event: string,
-    properties?: { [key: string]: any },
+    properties?: PostHogEventProperties,
```
marandaneto, line 322: I dont think we should allow `any` since its not serializable

posthog-core/src/index.ts
```diff
@@ -1060,7 +1065,6 @@ export abstract class PostHogCoreStateless {
       sentMessages.push(...batchMessages)
     }
     this._events.emit('flush', sentMessages)
-    return sentMessages
```
marandaneto, line 1063: not sure why we were returning `sentMessages` here

posthog-core/src/index.ts
```diff
@@ -929,7 +934,7 @@ export abstract class PostHogCoreStateless {
     })
   }
 
-  async flush(): Promise<any[]> {
+  async flush(): Promise<void> {
```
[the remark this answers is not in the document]
marandaneto: side effect of https://github.com/PostHog/posthog-js-lite/pull/493/files#r2073627712 i guess

posthog-core/src/index.ts
```diff
@@ -1447,19 +1451,19 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
       this.register({
         $groups: {
-          ...existingGroups,
+          ...(existingGroups as PostHogGroupProperties),
           ...groups,
         },
       })
 
-      if (Object.keys(groups).find((type) => existingGroups[type] !== groups[type])) {
+      if (Object.keys(groups).find((type) => existingGroups[type as keyof typeof existingGroups] !== groups[type])) {
```
marandaneto, line 1466: AI fix, have to test

posthog-core/src/types.ts
```diff
@@ -231,7 +234,7 @@ export type PostHogFlagsAndPayloadsResponse = Partial<
   Pick<PostHogDecideResponse, 'featureFlags' | 'featureFlagPayloads'>
 >
 
-export type JsonType = string | number | boolean | null | { [key: string]: JsonType } | Array<JsonType>
+export type JsonType = string | number | boolean | null | { [key: string]: JsonType } | Array<JsonType> | JsonType[]
```
marandaneto, line 237: sometimes only `Array<JsonType>` did not work and fixing everywhere was way more changes

posthog-core/src/index.ts
rafaeelaudibert: It's types, so if it passes CI, we're good

posthog-core/src/index.ts
```diff
@@ -1823,14 +1833,14 @@ export abstract class PostHogCore extends PostHogCoreStateless {
       this.capture('$feature_flag_called', {
         $feature_flag: key,
         $feature_flag_response: response,
-        $feature_flag_id: featureFlag?.metadata?.id,
-        $feature_flag_version: featureFlag?.metadata?.version,
-        $feature_flag_reason: featureFlag?.reason?.description ?? featureFlag?.reason?.code,
-        $feature_flag_bootstrapped_response: bootstrappedResponse,
-        $feature_flag_bootstrapped_payload: bootstrappedPayload,
+        $feature_flag_id: featureFlag?.metadata?.id ?? null,
```
rafaeelaudibert, line 1836: This changes behavior, right? Previously we'd not include the key when sending to the server, we now include the key with `null` value.

posthog-core/src/types.ts
rafaeelaudibert: I hate this whole `Array<T>` vs `T[]` thing, and I agree this is fine

posthog-core/src/index.ts
marandaneto: mm good point, not sure if we're dropping `key: undefined` or fetch would drop it, will test and fix if needed

posthog-core/test/posthog.flush.spec.ts
```diff
@@ -19,19 +19,15 @@ describe('PostHog Core', () => {
 
     it("doesn't fail when queue is empty", async () => {
       jest.useRealTimers()
-      await expect(posthog.flush()).resolves.toEqual([])
+      await expect(posthog.flush()).resolves.not.toThrow()
     })
 
     it('flush messages once called', async () => {
```
rafaeelaudibert, line 25: I guess this is why we were returning, it was useful for testing purposes. Can we assert we flushed all 3 elements here? Maybe inside the `mocks.fetch` expect?

rafaeelaudibert: See `successfulMessages` below

posthog-node/src/extensions/feature-flags/feature-flags.ts
```diff
@@ -470,19 +470,19 @@ class FeatureFlagsPoller {
 
         case 200: {
           // Process successful response
-          const responseJson = await res.json()
+          const responseJson = ((await res.json()) as { [key: string]: any }) ?? {}
```
rafaeelaudibert, line 473: JsonType could be used here, it's parsing JSON afterall

posthog-react-native/src/surveys/components/Surveys.tsx
```diff
@@ -47,8 +47,8 @@ export const dismissedSurveyEvent = (survey: Survey, posthog: PostHog): void =>
   posthog.capture('survey dismissed', {
     $survey_name: survey.name,
     $survey_id: survey.id,
-    $survey_iteration: survey.current_iteration,
-    $survey_iteration_start_date: survey.current_iteration_start_date,
+    $survey_iteration: survey.current_iteration ?? null,
```
rafaeelaudibert, line 50: Oh I really want a `?` operator to do this automatically

rafaeelaudibert (commented):
Got some comments, but looks good!

posthog-node/src/extensions/feature-flags/feature-flags.ts
marandaneto: thats what i did initially but then i'd need to cast to a key: value thing anyway otherwise `'flags' in responseJson` wouldn't work

posthog-react-native/src/surveys/components/Surveys.tsx
marandaneto: solved with a `maybeAdd` for now

posthog-core/src/types.ts
```diff
@@ -124,9 +128,8 @@ export type PostHogAutocaptureElement = {
   nth_child?: number
   nth_of_type?: number
   order?: number
-} & {
-  [key: string]: any
-} // Any key prefixed with `attr__` can be added
+} & PostHogEventProperties
+// Any key prefixed with `attr__` can be added
```
pauldambra, line 132: this comment is interesting, isn't it

there's nothing stopping us adding keys which aren't prefixed with `attr__` but that could be enforced by types here

not blocking on this PR just not seen that comment before

pauldambra (approved):
LGTM

posthog-core/src/types.ts
marandaneto: yeah comes from RN autocapture

pauldambra: so could have

```
type AttrPrefixed = {
  [key: `attr__${string}`]: JsonType
}
```

Issue #500, opened 2025-05-12
React native - "Promise.allSettled is not a function" - Expo SDK 53


candidosales:
## Bug description

I'm upgrading my project to use Expo SDK 53, and this issue is starting to rise:

```bash
 ERROR  Error while flushing PostHog [TypeError: Promise.allSettled is not a function (it is undefined)]
 WARN  Your app (or one of its dependencies) is using an outdated JSX transform. Update to the modern JSX transform for faster performance: https://react.dev/link/new-jsx-transform
 ERROR  Error while flushing PostHog [TypeError: Promise.allSettled is not a function (it is undefined)]
 ERROR  Error while flushing PostHog [TypeError: Promise.allSettled is not a function (it is undefined)]
```

My package.json:

```json
{
"dependencies": {
        "@clerk/clerk-expo": "2.11.3",
        "@clerk/types": "4.58.0",
        "@expo-google-fonts/fraunces": "^0.3.1",
        "@expo-google-fonts/inter": "^0.3.0",
        "@expo-google-fonts/lato": "^0.3.0",
        "@expo-google-fonts/poppins": "^0.3.1",
        "@expo/vector-icons": "^14.1.0",
        "@react-native-community/cli": "18.0.0",
        "@react-navigation/native": "^7.1.9",
        "@sentry/react-native": "~6.13.1",
        "@shopify/flash-list": "1.8.0",
        "@tanstack/react-query": "5.75.7",
        "expo": "53.0.9",
        "expo-application": "~6.1.4",
        "expo-blur": "~14.1.4",
        "expo-build-properties": "~0.14.6",
        "expo-constants": "~17.1.6",
        "expo-crypto": "~14.1.4",
        "expo-dev-client": "~5.1.8",
        "expo-device": "~7.1.4",
        "expo-file-system": "~18.1.10",
        "expo-font": "~13.3.1",
        "expo-haptics": "~14.1.4",
        "expo-linear-gradient": "~14.1.4",
        "expo-linking": "~7.1.5",
        "expo-localization": "~16.1.5",
        "expo-router": "~5.0.6",
        "expo-secure-store": "~14.2.3",
        "expo-splash-screen": "~0.30.8",
        "expo-status-bar": "~2.2.3",
        "expo-system-ui": "~5.0.7",
        "expo-updates": "~0.28.13",
        "expo-web-browser": "~14.1.6",
        "i18n-js": "4.5.1",
        "posthog-react-native": "^3.15.2",
[...]

PR #502, opened 2025-05-14
fix: use custom allSettled implementation to avoid issues with patching Promise

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js-lite/issues/500

### Bump level

- [ ] Major
- [ ] Minor
- [X] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [X] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/utils.ts
+++ b/posthog-core/src/utils.ts
@@ -196,9 +196,9 @@ export function allSettled<T>(
 ): Promise<({ status: 'fulfilled'; value: T } | { status: 'rejected'; reason: any })[]> {
   return Promise.all(
     promises.map((p) =>
-      (p ?? Promise.resolved()).then(
-        (value) => ({ status: 'fulfilled' as const, value }),
-        (reason) => ({ status: 'rejected' as const, reason })
+      (p ?? Promise.resolve()).then(
+        (value: any) => ({ status: 'fulfilled' as const, value }),
+        (reason: any) => ({ status: 'rejected' as const, reason })
       )
     )
   )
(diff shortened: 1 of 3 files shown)


robbie-c (approved):
Wow, I'm surprised that RN doesn't support allSettled

marandaneto:
> Reference in ne

it does after version X but libs/expo (not sure) are patching things and something is broken for a few folks

posthog-core/src/utils.ts
```diff
@@ -190,3 +190,16 @@ export function isTokenInRollout(token: string, percentage: number = 0, excluded
...
+): Promise<({ status: 'fulfilled'; value: T } | { status: 'rejected'; reason: any })[]> {
+  return Promise.all(
+    promises.map((p) =>
+      p.then(
+        (value) => ({ status: 'fulfilled' as const, value }),
+        (reason) => ({ status: 'rejected' as const, reason })
+      )
+    )
+  )
+}
```
robbie-c, line 205:
```suggestion
export function allSettled<T>(
  promises: (Promise<T> | null | undefined)[]
): Promise<({ status: 'fulfilled'; value: T } | { status: 'rejected'; reason: any })[]> {
  return Promise.all(
    promises.map((p) =>
      (p ?? Promise.resolved()).then(
        (value) => ({ status: 'fulfilled' as const, value }),
        (reason) => ({ status: 'rejected' as const, reason })
      )
    )
  )
}
```

marandaneto:
@robbie-c @rafaeelaudibert https://github.com/getsentry/sentry-react-native/issues/3957
sentry does the polyfill to capture rejected promises with a promise version that does not have the allSettled func

PR #504, opened 2025-05-20
fix: session recording respects linked flags

marandaneto:
## Problem

https://posthog.slack.com/archives/C04MZFDA5KK/p1746644228860679
bug introduced by the [migration of the flags endpoint](https://github.com/PostHog/posthog-js-lite/pull/458/)

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [x] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -1740,7 +1740,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
           // Mark that we hit the /decide endpoint so we can capture this in the $feature_flag_called event
           this.setPersistedProperty(PostHogPersistedProperty.DecideEndpointWasHit, true)
 
-          this.cacheSessionReplay('decide', res)
+          this.cacheSessionReplay('decide/flags', res)
         }
         return res
       })
--- a/posthog-core/src/index.ts
+++ b/posthog-core/src/index.ts
@@ -1600,7 +1600,9 @@ export abstract class PostHogCore extends PostHogCoreStateless {
     const sessionReplay = response?.sessionRecording
     if (sessionReplay) {
       this.setPersistedProperty(PostHogPersistedProperty.SessionReplay, sessionReplay)
-      this.logMsgIfDebug(() => console.log('PostHog Debug', `Session replay config from ${source}: `, JSON.stringify(sessionReplay)))
+      this.logMsgIfDebug(() =>
+        console.log('PostHog Debug', `Session replay config from ${source}: `, JSON.stringify(sessionReplay))
+      )
     } else if (typeof sessionReplay === 'boolean' && sessionReplay === false) {
       // if session replay is disabled, we don't need to cache it
       // we need to check for this because the response might be undefined (/flags does not return sessionRecording yet)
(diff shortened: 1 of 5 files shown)


posthog-core/src/index.ts
```diff
@@ -1593,13 +1596,15 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
-  private cacheSessionReplay(response?: PostHogRemoteConfig): void {
+  private cacheSessionReplay(source: string, response?: PostHogRemoteConfig): void {
     const sessionReplay = response?.sessionRecording
     if (sessionReplay) {
       this.setPersistedProperty(PostHogPersistedProperty.SessionReplay, sessionReplay)
-      this.logMsgIfDebug(() => console.log('PostHog Debug', 'Session replay config: ', JSON.stringify(sessionReplay)))
-    } else {
-      this.logMsgIfDebug(() => console.info('PostHog Debug', 'Session replay config disabled.'))
+      this.logMsgIfDebug(() => console.log('PostHog Debug', `Session replay config from ${source}: `, JSON.stringify(sessionReplay)))
+    } else if (typeof sessionReplay === 'boolean' && sessionReplay === false) {
```
marandaneto, line 1604: we only clear the data if sessionReplay=false because /flags does not return `sessionRecording` yet otherwise its gonna overwrite the cached value from the remote config
/decide?=v4 and remote config do return the sessionRecording config

posthog-core/src/index.ts
```diff
@@ -1773,7 +1779,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
     ) as PostHogFeatureFlagDetails
   }
 
-  private getKnownFeatureFlags(): PostHogDecideResponse['featureFlags'] | undefined {
+  protected getKnownFeatureFlags(): PostHogDecideResponse['featureFlags'] | undefined {
```
marandaneto, line 1782: easier to reuse this method that does all normalization otherwise I'd need to duplicate some code on the react native side or expose way more internal methods to the react native package

posthog-react-native/src/posthog-rn.ts
```diff
@@ -250,10 +251,10 @@ export class PostHog extends PostHogCore {
       if (OptionalReactNativeSessionReplay) {
         try {
           this._resetSessionId(OptionalReactNativeSessionReplay, sessionId)
-          this.logMsgIfDebug(() => console.info('PostHog Debug', `Session replay started with sessionId ${sessionId}.`))
+          this.logMsgIfDebug(() => console.info('PostHog Debug', `sessionId rotated from ${this._currentSessionId} to ${sessionId}.`))
```
marandaneto, line 254: the session id rotation can happen even without session replay so the log should be session id generic

posthog-core/src/index.ts
```diff
@@ -1622,36 +1629,39 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
-                console.log('PostHog Debug', 'Surveys fetched from remote config: ', JSON.stringify(surveys))
-              )
-            }
-
-            if (this.disableSurveys === false && hasSurveys) {
-              this.setPersistedProperty<SurveyResponse['surveys']>(
-                PostHogPersistedProperty.Surveys,
-                surveys as Survey[]
-              )
+            if (this.disableSurveys === false) {
```
marandaneto, line 1632: just to make logging cleaner if surveys is disabled

posthog-core/src/index.ts
pauldambra: 🙌

pauldambra (approved):

posthog-core/src/index.ts
```diff
@@ -1593,13 +1596,17 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
     const sessionReplay = response?.sessionRecording
     if (sessionReplay) {
       this.setPersistedProperty(PostHogPersistedProperty.SessionReplay, sessionReplay)
-      this.logMsgIfDebug(() => console.log('PostHog Debug', 'Session replay config: ', JSON.stringify(sessionReplay)))
-    } else {
-      this.logMsgIfDebug(() => console.info('PostHog Debug', 'Session replay config disabled.'))
+      this.logMsgIfDebug(() =>
+        console.log('PostHog Debug', `Session replay config from ${source}: `, JSON.stringify(sessionReplay))
+      )
+    } else if (typeof sessionReplay === 'boolean' && sessionReplay === false) {
```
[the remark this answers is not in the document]
marandaneto: I will add a test for this now

posthog-react-native/src/posthog-rn.ts
```diff
@@ -429,7 +455,12 @@ export class PostHog extends PostHogCore {
 
         try {
           if (!(await OptionalReactNativeSessionReplay.isEnabled())) {
-            await OptionalReactNativeSessionReplay.start(sessionId, sdkOptions, sdkReplayConfig, decideReplayConfig)
+            await OptionalReactNativeSessionReplay.start(
```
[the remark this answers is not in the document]
marandaneto: yes, that's how it is and we'll need to change it to support the manual start/stop.
the idea was to not pay the price of installing session replay at all if its disabled.

posthog-core/src/index.ts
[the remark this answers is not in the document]
marandaneto: `sessionReplay` can be a boolean or an object, so the first check will be true if `sessionReplay=true` which does not exist in our API contract or if `sessionReplay` is not null/undefined (an object) meaning enabled.
and `sessionReplay=false` disabled.
sessionReplay=undefined\null means the API does not return anything so we don't know if its enabled or not, we just leave it as it is

PR #514, opened 2025-05-29
feat: gemini support

k11kirky:
## Problem

Adding support for @google/genai to LLM observability

### DIFF
--- a/posthog-ai/package.json
+++ b/posthog-ai/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@posthog/ai",
-  "version": "4.5.0",
+  "version": "5.0.0",
   "description": "PostHog Node.js AI integrations",
   "repository": {
     "type": "git",
@@ -26,6 +26,9 @@
     "llm",
     "observability"
   ],
+  "engines": {
+    "node": ">=20.0.0"
+  },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.36.3",
     "@google/genai": "^1.1.0",
--- a/posthog-ai/package.json
+++ b/posthog-ai/package.json
@@ -22,11 +22,13 @@
     "ai",
     "openai",
     "anthropic",
+    "gemini",
     "llm",
     "observability"
   ],
   "dependencies": {
     "@anthropic-ai/sdk": "^0.36.3",
+    "@google/genai": "^1.1.0",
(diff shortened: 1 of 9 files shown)


posthog-ai/package.json
```diff
@@ -22,11 +22,13 @@
...
+    "gemini",
     "llm",
     "observability"
   ],
   "dependencies": {
     "@anthropic-ai/sdk": "^0.36.3",
+    "@google/genai": "^1.1.0",
     "@langchain/core": "^0.3.37",
     "ai": "^4.1.0",
     "langchain": "^0.3.15",
```
rafaeelaudibert, line 34: Random question: does it make sense to *require* these as dependencies? Shouldnt they be peer dependencies instead?

posthog-ai/src/index.ts
```diff
@@ -2,10 +2,12 @@ import PostHogOpenAI from './openai'
 import PostHogAzureOpenAI from './openai/azure'
 import { wrapVercelLanguageModel } from './vercel/middleware'
 import PostHogAnthropic from './anthropic'
+import PostHogGemini from './gemini'
 import { LangChainCallbackHandler } from './langchain/callbacks'
 
 export { PostHogOpenAI as OpenAI }
 export { PostHogAzureOpenAI as AzureOpenAI }
 export { PostHogAnthropic as Anthropic }
+export { PostHogGemini as GoogleGenAI }
```
rafaeelaudibert, line 11: Why not call it `PostHogGoogleGenAI` to keep it consistent?

rafaeelaudibert (approved):

posthog-ai/package.json
k11kirky: Good point. But will be a breaking change. I'll fix CI, merge this and do a follow up and bump the major Version

rafaeelaudibert: Yeah, good call

rafaeelaudibert:
@k11kirky, it looks like you won't be able to merge this anyways because of the node 18 -> 20 dependency requirement because of googleai

image

rafaeelaudibert (approved):
Like discussed in Slack:

- If you wanna be slightly nicer with our customers, @google/genai v1.0.0 supports Node 18. You could release a v4 with that, and then v5 with 1.3+

PR #520, opened 2025-06-05
chore: remove deprecated methods and options and rename stale props

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js-lite/issues/212

### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/src/posthog-rn.ts
+++ b/posthog-react-native/src/posthog-rn.ts
@@ -41,11 +41,9 @@ export type PostHogOptions = PostHogCoreOptions & {
 
   /** Captures app lifecycle events such as Application Installed, Application Updated, Application Opened, Application Became Active and Application Backgrounded.
    * By default is false.
-   * If you're already using the 'captureLifecycleEvents' options with 'withReactNativeNavigation' or 'PostHogProvider, you don't need to set this.
-   * If this is set, this value has priority over the 'captureLifecycleEvents' option.
    * Application Installed and Application Updated events are not supported with persistence set to 'memory'.
    */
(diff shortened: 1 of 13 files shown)


posthog-react-native/src/posthog-rn.ts
```diff
@@ -448,15 +445,14 @@ export class PostHog extends PostHogCore {
...
+  private async captureAppLifecycleEvents(): Promise<void> {
     const appBuild = this._appProperties.$app_build
     const appVersion = this._appProperties.$app_version
 
     const isMemoryPersistence = this._persistence === 'memory'
 
-    // version and build are deprecated, but we keep them for compatibility
-    // use $app_version and $app_build instead
-    const properties: PostHogEventProperties = { version: appVersion, build: appBuild }
+    // $app_version and $app_build are already added in the common event properties
```
rafaeelaudibert, line 454: Not sure we need this comment here

rafaeelaudibert (approved):

posthog-react-native/src/posthog-rn.ts
marandaneto: Since we add the previous version and build, you might think but whats about the current version and build? Well turns out its added elsewhere.

marandaneto: Good point, can do

PR #529, opened 2025-06-16
fix(flags): Handle double-encoded JSON response

haacked:
## Problem

See https://posthoghelp.zendesk.com/agent/tickets/32276 for the issue.

## Changes

If the response we get from the endpoint is a string, we try parsing as JSON again.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native

### Changelog notes

- fix: Handle double-encoded JSON payloads from the remote config endpoint

### DIFF
--- a/posthog-node/CHANGELOG.md
+++ b/posthog-node/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Next
 
+# 5.1.1 - 2025-06-16
+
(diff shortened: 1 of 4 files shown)


dmarticus (approved):
cool

PR #548, opened 2025-06-26
fix: multiple choice response

ioannisj:
## Problem

see: https://posthog.com/questions/issues-with-post-hog-surveys-in-react-native

- Add disabled style when the survey button is disabled
- Fixed multiple choice response to always be a string array

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [x] posthog-react-native
- [ ] posthog-nextjs-config

### Changelog notes

- Added support for X

### DIFF
--- a/posthog-react-native/src/surveys/components/QuestionTypes.tsx
+++ b/posthog-react-native/src/surveys/components/QuestionTypes.tsx
@@ -257,8 +257,9 @@ export function MultipleChoiceQuestion({
         onSubmit={() => {
           // If open choice is selected, replace the choice name with the actual value entered
           const result = selectedChoices.map((c) => (c === openChoice ? openEndedInput : c))
-          // Always submit as an array, even when there's only one selection
-          onSubmit(result)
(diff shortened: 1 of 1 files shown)


posthog-react-native/src/surveys/components/QuestionTypes.tsx
```diff
@@ -257,7 +257,8 @@ export function MultipleChoiceQuestion({
         onSubmit={() => {
           // If open choice is selected, replace the choice name with the actual value entered
           const result = selectedChoices.map((c) => (c === openChoice ? openEndedInput : c))
-          onSubmit(result.length === 1 ? result[0] : result)
+          // Always submit as an array, even when there's only one selection
+          onSubmit(result)
```
marandaneto, line 261: for single choice questions, will that be an array as well? i think that would be wrong iirc

PR #559, opened 2025-07-07
feat(flags): decouple local evaluation from personal API keys; support decrypting remote config payloads without relying on the feature flags poller

dmarticus:
## Problem

We have this problem with our server-side SDKs where we start the feature flag poller (for local evaluation) as soon as customers supply a `personalApiKey`.  This led to reports (e.g.: https://posthog.slack.com/archives/C04J1TJ11UZ/p1751499805232469) where customers were mentioning not wanting to pay for local eval when their goal for supplying the personal API key was to just use the `getRemoteConfigPayload` method.

Coupling the personal API key to feature flag polling is an artifact of when the only reason to supply a personal API key to the posthog-node library was to do local eval, but now that it's used for at least remote config decryption (and possibly more things), I think it makes sense to decouple those components.

## Changes

- decoupled `getRemoteConfigPayload` from the feature flag poller
- added an optional `enableLocalEvaluation` parameter to the posthog config that defaults to `true` when the `personalApiKey` is provided, but can be overridden by users who want to provide an API but not use local evaluation.  Designing it this way means that folks can safely upgrade their posthog-node version without changing any of the configuration if they wanted the coupled behavior to still exist.
- updated tests

### Bump level

- [ ] Major
- [x] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native
- [ ] posthog-nextjs-config

### Changelog notes

- feat(flags): decouple local evaluation from personal API keys; support decrypting remote config payloads without relying on the feature flags poller

### DIFF
--- a/posthog-node/src/client.ts
+++ b/posthog-node/src/client.ts
@@ -8,6 +8,7 @@ import {
   PostHogFetchResponse,
   PostHogFlagsAndPayloadsResponse,
   PostHogPersistedProperty,
+  safeSetTimeout,
 } from 'posthog-core'
 import { EventMessage, GroupIdentifyMessage, IdentifyMessage, IPostHog, PostHogOptions } from './types'
 import { FeatureFlagDetail, FeatureFlagValue } from 'posthog-core'
@@ -667,7 +668,7 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     let abortTimeout = null
     if (this.options.requestTimeout && typeof this.options.requestTimeout === 'number') {
       const controller = new AbortController()
-      abortTimeout = setTimeout(() => {
+      abortTimeout = safeSetTimeout(() => {
         controller.abort()
       }, this.options.requestTimeout)
       options.signal = controller.signal
--- a/posthog-node/src/client.ts
+++ b/posthog-node/src/client.ts
@@ -648,10 +648,6 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     return super._shutdown(shutdownTimeoutMs)
   }
 
-  /**
-   * @param flagKey
-   * @returns
-   */
   private async _requestRemoteConfigPayload(flagKey: string): Promise<PostHogFetchResponse | undefined> {
     if (!this.options.personalApiKey) {
       return undefined
--- a/posthog-node/src/client.ts
+++ b/posthog-node/src/client.ts
@@ -50,21 +50,26 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
         )
       }
 
-      this.featureFlagsPoller = new FeatureFlagsPoller({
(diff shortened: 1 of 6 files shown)


posthog-node/src/client.ts
```diff
@@ -639,6 +648,43 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+        ...this.getCustomHeaders(),
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${this.options.personalApiKey}`,
+      },
+    }
+
+    let abortTimeout = null
+    if (this.options.requestTimeout && typeof this.options.requestTimeout === 'number') {
+      const controller = new AbortController()
+      abortTimeout = setTimeout(() => {
```
haacked, line 670: In the old implementation this was `safeSetTimeout`. Is there a reason it was changed to `setTimeout` here? Seems like `safeSetTimeout` is the right choice.

```suggestion
      abortTimeout = safeSetTimeout(() => {
```

posthog-node/src/client.ts
```diff
@@ -639,6 +648,43 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+    const url = `${this.host}/api/projects/@current/feature_flags/${flagKey}/remote_config/`
+
+    const options: PostHogFetchOptions = {
+      method: 'GET',
+      headers: {
+        ...this.getCustomHeaders(),
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${this.options.personalApiKey}`,
+      },
+    }
```
haacked, line 665: Why not use `getPersonalApiKeyRequestOptions` here like before?

posthog-node/src/types.ts
```diff
@@ -57,6 +57,9 @@ export type PostHogOptions = PostHogCoreOptions & {
   // Maximum size of cache that deduplicates $feature_flag_called calls per user.
   maxCacheSize?: number
   fetch?: (url: string, options: PostHogFetchOptions) => Promise<PostHogFetchResponse>
+  // Whether to enable feature flag polling for local evaluation by default. Defaults to true when personalApiKey is provided.
+  // We recommend setting this to false if you are using the personalApiKey for evaluating remote config payloads via `getRemoteConfigPayload`.
```
haacked, line 61:
```suggestion
  // We recommend setting this to false if you are only using the personalApiKey for evaluating remote config payloads via `getRemoteConfigPayload` and not using local evaluation.
```

haacked (approved):
Nice. Some suggested changes and nits, but they're pretty minor and nothing blocking.

posthog-node/src/client.ts
dmarticus: ehhh, it was localize the the `FeatureFlagsPoller` class and I couldn't be bothered to factor it out.

PR #564, opened 2025-07-09
docs: Add CHANGELOG notice about GZip compression

rafaeelaudibert:
PostHog web now attempts to compress messages with GZip before sending them to our servers if available in your runtime. This shouldn't cause you any problems, but if someone's tests are attempting to assert on the data that's sent to our servers, they'll now need to decompress it first.

Let's note this on the BREAKING CHANGES section of both our web and node versions

See https://github.com/PostHog/posthog-js-lite/pull/481

### DIFF
--- a/posthog-node/CHANGELOG.md
+++ b/posthog-node/CHANGELOG.md
@@ -27,6 +27,7 @@
 ## Breaking changes
 
 1. feat: migrate to native fetch, Node 20+ required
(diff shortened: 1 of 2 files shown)


posthog-node/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 ## Breaking changes
 
 1. feat: migrate to native fetch, Node 20+ required
+2. PostHog Node now compresses messages with GZip before sending them to our servers when the runtime supports compression. Network traffic interceptors and test assertions on payloads must handle GZip decompression to inspect the data
```
clemens, line 30: Alternatively, one can set `{ disableCompression: true }` in the client instantiated in the tests. (That's how we're doing it.)

PR #573, opened 2025-07-15
fix: use correct types for PostHogContext which cascades through usePostHog hook and callers

cbarber:
## Problem

We discovered that usePostHog was typed incorrectly:
```
diff --git a/node_modules/posthog-react-native/lib/posthog-react-native/src/hooks/usePostHog.d.ts b/node_modules/posthog-react-native/lib/posthog-react-native/src/hooks/usePostHog.d.ts
index 4268052..e1ab864 100644
--- a/node_modules/posthog-react-native/lib/posthog-react-native/src/hooks/usePostHog.d.ts
+++ b/node_modules/posthog-react-native/lib/posthog-react-native/src/hooks/usePostHog.d.ts
@@ -1,2 +1,2 @@
 import { PostHog } from '../posthog-rn';
-export declare const usePostHog: () => PostHog;
+export declare const usePostHog: () => PostHog | undefined;
```

Checking PostHogContext, we can see that the client type is initialized as `undefined` and then coerced into PostHog:
https://github.com/PostHog/posthog-js-lite/blob/0b983d8305a0b20fb0940bbbb43195de9b7ca2e4/posthog-react-native/src/PostHogContext.ts#L4C103-L4C113

## Changes

Properly type the context and usePostHog hook, and then slap our troubles away with the optional chaining operator. I don't think this is the correct solution, but don't have enough context for the other features. It is at least an improvement of the call onto a maybe undefined. Happy to fix up with a different approach. 

### Bump level

- [x] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [x] posthog-react-native
- [ ] posthog-nextjs-config

### Changelog notes

- Added correct type for `usePostHog`

### DIFF
--- a/posthog-react-native/src/PostHogContext.ts
+++ b/posthog-react-native/src/PostHogContext.ts
@@ -1,4 +1,4 @@
 import React from 'react'
 import { PostHog } from './posthog-rn'
 
-export const PostHogContext = React.createContext<{ client: PostHog }>({ client: undefined as unknown as PostHog })
+export const PostHogContext = React.createContext<{ client: PostHog | undefined }>({ client: undefined })
--- a/posthog-react-native/src/hooks/useFeatureFlag.ts
+++ b/posthog-react-native/src/hooks/useFeatureFlag.ts
@@ -7,12 +7,12 @@ export function useFeatureFlag(flag: string, client?: PostHog): FeatureFlagValue
   const contextClient = usePostHog()
   const posthog = client || contextClient
 
-  const [featureFlag, setFeatureFlag] = useState<FeatureFlagValue | undefined>(posthog.getFeatureFlag(flag))
+  const [featureFlag, setFeatureFlag] = useState<FeatureFlagValue | undefined>(posthog?.getFeatureFlag(flag))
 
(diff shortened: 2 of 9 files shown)


cbarber:
Yes, so we have multiple environments and a toggle in our application for developers to switch between then. In `development` mode, the posthogApiKey is left as `undefined`:
```
      <PostHogProvider
        apiKey={posthogApiKey ?? 'disabled'}
        options={{
          host: 'https://us.i.posthog.com',
          disabled: !posthogApiKey,
        }}>
        {children}
      </PostHogProvider>
```

This results in an `undefined` returned from `usePostHog` which typescript will not enforce is safely accessed. A NoOp would work here and satisfy the typing of always having a `PostHog` instance. 

I'd recommend keeping testing for `undefined` in the hook to throw a message calling out the configuration. Let me take another pass.

marandaneto:
yes, what David said, please open a new PR based on https://github.com/PostHog/posthog-js/tree/main/packages/react-native, thanks

Issue #535, opened 2025-06-19
TypeError: Class extends value undefined is not a constructor or null when using @posthog/ai with openai v5.1.0


RaminGe:
## Bug description

When using `@posthog/ai` version 5.1.0 together with the latest `openai` package (`v5.1.0` or `v5.1.1`), the following runtime error occurs:
```bash
/Users/XXXX/node_modules/@posthog/ai/src/openai/index.ts:40
export class WrappedChat extends OpenAIOrignal.Chat {
                                               ^


TypeError: Class extends value undefined is not a constructor or null
    at Object.<anonymous> (/Users/XXXX/node_modules/@posthog/ai/src/openai/index.ts:40:48)
    at Module._compile (node:internal/modules/cjs/loader:1734:14)
    at Object..js (node:internal/modules/cjs/loader:1899:10)
    at Module.load (node:internal/modules/cjs/loader:1469:32)
    at Function._load (node:internal/modules/cjs/loader:1286:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
    at Module.require (node:internal/modules/cjs/loader:1491:12)
    at require (node:internal/modules/helpers:135:16)
    at Object.<anonymous> (/Users/XXXX/ai.service.ts:3:34)

Node.js v23.11.0
```

## How to reproduce

1. Install the following package versions:
   - `openai@5.1.0`
   - `@posthog/ai@5.1.0`
2. Use `@posthog/ai` in your code (e.g. import and create `new OpenAI({})`)
3. Run the project

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-ai
- [ ] posthog-react-native
- [ ] posthog-nextjs

## Additional context
My environment:
- Node.js version: `v23.11.0`
- OS: macOS
- Installed using `npm`

k11kirky:
Hey @RaminGe I have been unable to replicate this issue with the version listed. Can you try rm -rf node_modules and
rm package-lock.json then after a npm install check the version of the modules you have installed inside your node_modules folder

Radu-Raicea:
Hey @RaminGe and @RP-9!

This issue should now be fixed in 5.2.1.

Thank you for sharing both the minimal setup and your observation about the changes being impacted by the move from v4 to v5! 🙌

PR #580, opened 2025-07-18
fix: avoid undefined errors when extending OpenAI classes

Radu-Raicea:
## Problem

Consumers of `@posthog/ai` encountered runtime crashes when using it alongside OpenAI SDK v5.x.

The issue was that `@posthog/ai` internally extended OpenAI classes like `Chat`, `Completions`, and `Responses` at module load time. However, in OpenAI SDK v5, these classes are only assigned to the `OpenAI` object at runtime — they are not immediately available as static properties.

This caused errors like:

```
TypeError: Class extends value undefined is not a constructor or null
```

…whenever someone tried to import and instantiate the OpenAI wrapper from `@posthog/ai`.

Issue: #535.

## Changes

- Deferred access to `Chat`, `Completions`, and `Responses` by assigning them to local constants after the OpenAI SDK has finished its runtime setup.
- Replaced all `OpenAI.Chat`, `OpenAI.Chat.Completions`, and `OpenAI.Responses` references used during class extension.
- This ensures the resource classes are defined before subclassing them, preventing `undefined` runtime errors.

### Bump level

- [ ] Major
- [ ] Minor
- [x] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-ai
- [ ] posthog-react-native
- [ ] posthog-nextjs-config

### Changelog notes

- Fixed crash when importing `@posthog/ai` with OpenAI SDK v5.x by deferring access to Chat, Completions, and Responses classes until runtime.

### DIFF
--- a/posthog-ai/src/openai/index.ts
+++ b/posthog-ai/src/openai/index.ts
@@ -1,10 +1,14 @@
-import OpenAIOrignal, { ClientOptions } from 'openai'
+import { OpenAI as OpenAIOrignal, ClientOptions } from 'openai'
 import { PostHog } from 'posthog-node'
 import { v4 as uuidv4 } from 'uuid'
 import { formatResponseOpenAI, MonitoringParams, sendEventToPosthog } from '../utils'
 import type { APIPromise } from 'openai'
 import type { Stream } from 'openai/streaming'
 import type { ParsedResponse } from 'openai/resources/responses/responses'
+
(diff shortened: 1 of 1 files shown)


k11kirky:
Looks good - Just need to bump the version in posthog/ai and add the fix to the change log

k11kirky (approved):
Nice work

PR #591, opened 2025-07-24
chore: simplify access to NPM_TOKEN secret

Piccirello:
## Problem

All branches now use the same token, so this can be simplified. It seems the TODO also no longer applies.

### Bump level

- [ ] Major
- [ ] Minor
- [ ] Patch

### Libraries affected

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [ ] posthog-ai
- [ ] posthog-react-native
- [ ] posthog-nextjs-config

### Changelog notes

- Added support for X

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,15 +13,10 @@ jobs:
       matrix:
         package:
           - name: posthog-react-native
-            npm_token_secret: NPM_TOKEN # TODO: Change to standard token once able
           - name: posthog-node
-            npm_token_secret: NPM_TOKEN # TODO: Change to standard token once able
           - name: posthog-web
-            npm_token_secret: NPM_TOKEN # TODO: Change to standard token once able
           - name: posthog-ai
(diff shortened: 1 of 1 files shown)


marandaneto (approved):

marandaneto:
> @Piccirello we're in the process of merging the `posthog-js-lite` repo into `posthog-js`. I think this has already been improved in
> 
> [PostHog/posthog-js@`afeec0a`/.github/actions/release/action.yaml#L48](https://github.com/PostHog/posthog-js/blob/afeec0a268e76d413a50bff42e25fbe2a1b50162/.github/actions/release/action.yaml#L48)

there was an issue on the JS repo with tokens so releases were blocked, is that fixed @daibhin ? otherwise i'd say we could merge this in since it wont require changes on the JS SDK anyway

daibhin:
I believe it's fixed and everything is working as expected on `posthog-js` repo. @hpouillot might know better

marandaneto:
> I believe it's fixed and everything is working as expected on `posthog-js` repo. @hpouillot might know better

it is all fixed already!

